PR #3, opened 2016-11-23
Add section for phantom types and GADTs

cdparks:


haskell-best-practices.md
```diff
@@ -184,6 +184,165 @@ instance Monoid Mult where
...
+
+data PasswordState = Raw | Hashed
+
+newtype Password (s :: PasswordState) = Password Text
+
+rawPassword :: Text -> Password 'Raw
+rawPassword = Password
+
+hashPassword :: Password 'Raw -> Password 'Hashed
+hashPassword (Password r) = Password (hash r)
```
akurilin: might want to clarify somewhere where `hash` comes from so people don't go around looking for it?

cdparks:
Yeah, I was referring to some abstract `hash` function not a real one

haskell-best-practices.md
```diff
@@ -184,6 +184,172 @@ instance Monoid Mult where
...
+  VI y' <- eval y
+  return $ VB $ x' < y'
+eval (Cond c t f) =
+  VB c' <- eval c
+  if c'
+    then eval t
+    else eval f
+```
+
+We're relying on the `Monad` instance for for `Maybe` to call `fail` on pattern match failures when we pass something like `Add (I 1) (B True)`. We also have to make an extra datatype to represent values.
```
akurilin: double for

akurilin: Also think focus should be on pointing out that this is a run-time exception vs something we can detect at compilation

haskell-best-practices.md
```diff
@@ -184,6 +184,172 @@ instance Monoid Mult where
...
+When we pattern match on `B` here, we gain access to evidence that `b` is a `Bool`, and so on for the other constructors. GHC won't even let us construct a value like `Add (I 1) (B True)`:
+```haskell
+Couldn't match type ‘Bool’ with ‘Int’
+ Expected type: Expr Int
+   Actual type: Expr Bool
+In the second argument of ‘Add’, namely ‘(B True)’
+In the expression: Add (I 1) (B True)
+```
+
+If you're curious how this works, you can dump the core (`-ddump-simpl`) to see that each constructor is literally carrying around an extra parameter as evidence that allows GHC to insert safe type casts from `a` to `Int` or `Bool` (or whatever) inside a pattern match.
```
akurilin: Can add links to Core in the appendix if people are wondering what that is

akurilin (changes requested):

haskell-best-practices.md
cdparks: Ah, good call

cdparks:
@AKurilin I think we'll eventually want a section on Core since it's useful for occasionally debugging performance and understanding how certain "magic" things work (e.g. typeclass dictionary elaboration).

akurilin:
kk!

PR #5, opened 2017-04-07
Create partial Guide for atomic component design

joris974:


pbrisbin:
Hey @joris974 this has been hanging around a while, do you think we should Close it?

FWIW, it is written with a lot of future tense about what *will* happen, not like our typical guides (which use *should* or *must* type language). Should this maybe be an RFC instead?

joris974:
This is a pattern that we have now. I may have write this to just have a place to clarify what was in my mind I don't think that is useful anymore

PR #7, opened 2017-09-15
Imports, exports, and extensions

pbrisbin:
These are aspects of my own style that I see *mostly* followed around our
codebase, so I'd like to float the idea of canonicalizing them.

Apologies if this is pre-trodden territory; I thought a PR would be an easy,
central, visible place to discuss these with the team. Happy to Close, or
change drastically, if that's where we end up.

In general, I'm of the opinion we should have *every style decision possible*
canonicalized (and ideally machine-checked), regardless of what that decision
is.

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -366,6 +366,27 @@ import qualified Data.Map as Map
 import qualified Data.Text as Text
 ```
 
+**NOTE**: if explicit imports exceed 80 columns, switch to a (sorted) list:
+
+```haskell
+-- Bad
+-- Long lines are hard to scan, inserting or removing an item is a noisier diff,
+-- and it's difficult to sort a horizontal list
+import System.IO (hPutStrLn, stderr, stdout, withFile, IOMode(..), hGetContents, hFlush, hClose)
+
+-- Good
+import System.IO
+  ( IOMode(..)
+  , hClose
+  , hFlush
+  , hGetContents
+  , hPutStrLn
+  , stderr
+  , stdout
+  , withFile
+  )
+```
+
 ### Importing types and qualifying
 It is also common to explicitly import types from a module and also import it qualified.
 ```
@@ -391,3 +412,134 @@ import qualified Data.List.NonEmpty as NE
 
 import qualified Data.Sequence as Seq
 ```
+
+### Import groups
+
+Imports should be divided into groups:
+
+1. Prelude (when explicit)
+1. Un-`qualified` dependency libraries
+1. `qualified` dependency libraries
+1. Un-`qualified` FrontRow libraries
+1. `qualified` FrontRow libraries
+1. Un-`qualified` modules from this library
+1. `qualified` modules from this libraries
+
+Aside from the empty line between groups, there should be no other whitespace.
+
+```haskell
+-- Bad
+-- What's coming from where?
+import ClassyPrelude
+import Network.AWS.S3
+import Network.AWS
+import Data.Text (Text)
+import Json
+import qualified Data.Text as T
+import Control.Lens
+import Unit
+import Data.Conduit
+import TextAssets.S3
+import Data.Set (Set)
+import qualified Data.Set as Set
+import Data.Map (Map)
+import qualified Data.Map as Map
+
+-- Bad
+-- You can't quickly highlight and :sort things
+import ClassyPrelude
+
+import Network.AWS.S3
+import Network.AWS
+import Data.Text (Text)
+import qualified Data.Text as T
+import Control.Lens
+import Data.Conduit
+import Data.Set (Set)
+import qualified Data.Set as Set
+import Data.Map (Map)
+import qualified Data.Map as Map
+
+import TextAssets.S3
+
+import Json (Json)
+import qualified Json as J
+import Unit
+
+-- Good
+import ClassyPrelude
+
+import Control.Lens
+import Data.Conduit
+import Data.Map (Map)
+import Data.Set (Set)
+import Data.Text (Text)
+import Network.AWS
+import Network.AWS.S3
+
+import qualified Data.Map as Map
+import qualified Data.Set as Set
+import qualified Data.Text as T
+
+import TextAssets.S3
+
+import Json (Json)
+import Unit
+
+import qualified Json as J
+```
+
+## Exports
+
+Use sorted, multi-line exports. There are two exceptions to this rule:
+
+1. A single-line export for `Main`, when it only exports `main`:
+
+   ```haskell
+   module Main (main) where
+   ```
+
+1. If the order of exports matters in your desired Haddock output, you may
+   violate sorting to achieve it.
+
+Otherwise:
+
+```haskell
+-- Bad
+module Driver (scienceOptions, socialStudiesOptions, mainWith) where
+
+-- Good
+module Driver
+  ( mainWith
+  , scienceOptions
+  , socialStudiesOptions
+  ) where
+```
+
+## Declaring Extensions
+
+- Declare extensions in the module(s) that use them, i.e. don't use
+  `default-extensions`
+
+  This ensures a) the source code works un-changed with other Haskell tooling
+  (e.g. `ghci`, `doctest`, etc) without further configuration and b) active
+  extensions are visible in the modules they impact.
+
+- Place extensions on their own line, and sort them
+
+  ```haskell
+  -- Bad
+  {-# LANGUAGE OverloadedStrings, RecordWildCards,
+      DataKinds #-}
+
+  -- Good
+  {-# LANGUAGE DataKinds #-}
+  {-# LANGUAGE OverloadedStrings #-}
+  {-# LANGUAGE RecordWildCards #-}
+  ```
+
+## Yesod
+
+When well-established Yesod idioms clash with the rules of this styleguide (e.g.
+the Yesod scaffold uses `default-extensions`), it is acceptable to follow the
+Yesod idioms and violate this style guide.
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -415,78 +415,59 @@ import qualified Data.Sequence as Seq
 
 ### Import groups
 
-Imports should be divided into groups:
+Put one blank line between `module`-`where` and the start of your `import`s. Put
+your preferred prelude (when explicit) first, followed by a blank line, then the
+rest of your imports.
 
-1. Prelude (when explicit)
-1. Un-`qualified` dependency libraries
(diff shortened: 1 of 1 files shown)


haskell-style.md
```diff
@@ -391,3 +412,134 @@ import qualified Data.List.NonEmpty as NE
...
+
+  -- Good
+  {-# LANGUAGE DataKinds #-}
+  {-# LANGUAGE OverloadedStrings #-}
+  {-# LANGUAGE RecordWildCards #-}
+  ```
+
+## Yesod
+
+When well-established Yesod idioms clash with the rules of this styleguide (e.g.
```
jdreaver: When does a well-established yesod idiom clash with the style guide? I thought we allowed `default-extensions`?

haskell-style.md
```diff
@@ -391,3 +412,134 @@ import qualified Data.List.NonEmpty as NE
 
 import qualified Data.Sequence as Seq
 ```
+
+### Import groups
+
+Imports should be divided into groups:
+
+1. Prelude (when explicit)
```
jdreaver: Separating into 3 groups is great, but we generally don't separate even further into qualified and unqualified. 

Also, we generally sort the imports in each group alphabetically.

jdreaver (changes requested):

haskell-style.md
```diff
@@ -391,3 +412,134 @@ import qualified Data.List.NonEmpty as NE
...
+
+-- Good
+module Driver
+  ( mainWith
+  , scienceOptions
+  , socialStudiesOptions
+  ) where
+```
+
+## Declaring Extensions
```
eborden, line 500: Yeah, I'm a huge fan of this. Especially since some extensions can clash and also can complicate errors and inference. What I usually see in practice is that orgs identify a subset of extensions to turn on by default ( `OverloadedStrings`, `LambdaCase`) and then all other extensions must be in a pragma.

haskell-style.md
pbrisbin: > Also, we generally sort the imports in each group alphabetically.

Yes, this patch includes that rule.

> we generally don't separate even further into qualified and unqualified.

Would folks be against starting to do that? Or is there a strict "let's canonicalize what we currently do (first)" rule at play? Not having this separation makes it harder to maintain the sort, which we agree is important.

haskell-style.md
pbrisbin: This patch adds the rule (with justification) to not use `default-extensions`, hence that being my clash example here.

The scaffold also generally does not follow our style, it does a lot of vertical alignment for example, and I wouldn't want to feel pressure to re-format the world (at least while we don't have a viable auto-formatter) after using the scaffold.

jdreaver: Ah I see what you mean now. Honestly I wouldn't want to codify this because it creates some ambiguity. Also, we can change the scaffold however we want at this point, generating it was a one-time thing.

haskell-style.md
```diff
@@ -366,6 +366,27 @@ import qualified Data.Map as Map
 import qualified Data.Text as Text
 ```
 
+**NOTE**: if explicit imports exceed 80 columns, switch to a (sorted) list:
```
eborden, line 369: :+1:

haskell-style.md
```diff
@@ -391,3 +412,134 @@ import qualified Data.List.NonEmpty as NE
...
+import Data.Map (Map)
+import qualified Data.Map as Map
+
+import TextAssets.S3
+
+import Json (Json)
+import qualified Json as J
+import Unit
+
+-- Good
```
eborden: While this looks clean and I've seen it in practice, I don't really see the benefit. A sorted list of imports with prudent use of explicit importing and qualification is easy to traverse. With this you end up having to maintain artificial segments of package types and all new hires must be indoctrinated. Sort your imports is low complexity with high yield.

haskell-style.md
```diff
@@ -391,3 +412,134 @@ import qualified Data.List.NonEmpty as NE
...
+
+import TextAssets.S3
+
+import Json (Json)
+import Unit
+
+import qualified Json as J
+```
+
+## Exports
```
eborden: :+1:

eborden (commented):
Only one real difference of opinion here.

haskell-style.md
jdreaver: > Would folks be against starting to do that? 

I personally would. It is common to import container modules qualified and unqualified

```haskell
import Data.Map (Map)
import qualified Data.Map as Map
```

I would rather not have these be split up because of this rule. Also, sometimes you have a diff where all you did in the import list was qualify an import, and with this rule it would move that import to a different line.

I know that `qualified` throws off naive line sorts, but maybe we should have a simple tool that sorts them instead then if we want to automate it (I usually just to a naive sort and then put the qualified modules where they should be, but I also don't remember the last time I've had to do this).

jdreaver (changes requested):

haskell-style.md
eborden: Agreed, screw yesod defaults. We aren't using the scaffold tool anyway.

haskell-style.md
eborden: Deleted my comment and moving it over here:

While this looks clean and I've seen it in practice, I don't really see the benefit. A sorted list of imports with prudent use of explicit importing and qualification is easy to traverse. With this you end up having to maintain artificial segments of package types and all new hires must be indoctrinated. Sort your imports is low complexity with high yield.

cbeav: Yeah, I proposed this once before and got pretty much the same response. The OCD in me likes to separated qualified vs. non-qualified, but the obsequious sycophant just wants everyone to be happy. Really glad we're finally writing this down somewhere.

eborden: How about this simple script?

```
#!/usr/bin/env stack
-- stack runghc --package text
{-# LANGUAGE OverloadedStrings #-}

import Data.List (sortBy)
import Data.Ord (comparing)
import Data.Text as T
import Data.Text.IO as T
import System.IO (stdin, stdout)

main = do
  input <- T.hGetContents stdin
  traverse (T.hPutStrLn stdout) . sortBy (comparing unprefixImport) $ T.lines input

unprefixImport x
  | T.null x = x
  | otherwise = T.replace " " ""  $ T.replace "qualified" "" $ T.replace "import" "" x
```

pbrisbin: Sure! but...

```hs
T.interact $ T.unlines . sortBy (comparing unprefixImport) . T.lines
```

eborden: Alternative script:
```
#!/bin/sh

Delimiter='\{\{\{'
ImportRex='(import\s*)(qualified\s*)?'
ToSortRex="s/^$ImportRex(.*)$/\\3$Delimiter\\1\\2/g"
FromSortRex="s/^([^$Delimiter]*)$Delimiter$ImportRex$/\\2\\3\\1/g"

sed -E "$ToSortRex" | sort | sed -E "$FromSortRex"
```

pbrisbin:
I think I'm sensing consensus. I've updated the import section.

jdreaver:
Ehhh, I think at least separating 3rd party imports and imports from our code is valuable. I actually agree with your 3 sections idea (3rd party, imports from other packages of ours, imports from the current package).

pbrisbin:
Yeah, I liked that too, but I took @eborden to mean 1 big list when we he said:

> A sorted list of imports with prudent use of explicit importing and qualification is easy to traverse. With this you end up having to maintain artificial segments of package types and all new hires must be indoctrinated. Sort your imports is low complexity with high yield.

So what say we all?

1. Prelude then one big list
1. Prelude, dependency libs, our libs, this lib

I can take or leave Prelude being separated, but I do prefer it first.

*EDIT*: I vote 2, in case that's not clear.

jdreaver:
I vote option 2

cdparks (approved):

eborden:
I prefer not to segment since I think these things often follow on their own by module naming. For example most all of our modules start with `FrontRow`, so they already cluster. However I won't fight city hall if y'all agree on it.

pbrisbin:
@cdparks Approved while the PR was in the state of Option 1, so I take that as a vote for that? If that's the case it's an even split so far.

eborden:
Get @AKurilin in here to break some ties if no one else has an opinion.

akurilin:
I'd say option 1 unless people feel they're losing productivity due to things not being as carefully arranged as option 2 would suggest.

pbrisbin:
With the tie broken, I believe this is merge-able as it is. @jdreaver that OK with you?

jdreaver:
Yeah LGTM

jdreaver (approved):

haskell-style.md
```diff
@@ -391,3 +412,109 @@ import qualified Data.List.NonEmpty as NE
...
 import qualified Data.Sequence as Seq
 ```
+
+### Import groups
+
+Put one blank line between `module`-`where` and the start of your `import`s. Put
+your preferred prelude (when explicit) first, followed by a blank line, then the
+rest of your imports.
+
+Imports are sorted, but with `qualified` versions appearing where they would
```
5outh, line 422: I'm a bit late here, but for import sorting I typically just run `:sort i` on the visual line block of imports in vim. That will group all `qualified` imports together, not interspersed like described here. Can we configure `stylish-haskell` or another code formatter to get this behavior automatically?

5outh (commented):
Overall this looks good to me (sorry fot the late review). Just had one comment about the order of `qualified` imports.

haskell-style.md
pbrisbin: > I typically just run :sort i on the visual line block of imports in vim. That will group all qualified imports together, not interspersed like described here

This is exactly why I dislike interspersing the `qualified` entries. Now I need to `:sort` and then manually move individual items to get it right.

I think this ship may have sailed, but I'm happy to participate in a discussion around a "one big group + simple sort (so `qualified`s fall together at the end)" option, if you'd want to re-spark it. I may have scared folks off by suggesting categorical grouping at the same time as `qualified`s at end, which made the whole thing feel like a more complex guide.

5outh: The other option is to use `stylish-haskell` with context-sensitive indenting turned off, which I think will group everything properly.

Then `:! stylish-haskell` over the import lists will work properly.

eborden: Woah, wait. You can turn off the indentation on `stylish-haskell`? I think that is the solution!

eborden: PR here: https://github.com/frontrowed/megarepo/pull/1549

PR #8, opened 2017-11-06
Update where and import-sorting guides

pbrisbin:
See distinct commits.

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -246,41 +246,36 @@ someBinding'' mx =
 ```
 
 #### Where bindings
+
+Expressions with `where`-bindings should place the body of the expression (if on
+the next line) and the `where`-bindings at the same (2-space) indentation, with
+the `where` keyword de-dented 1 space. Note that this causes an odd, 1-space
+indentation of the `where` keyword.
+
 ```haskell
--- We're pretty liberal with where bindings, but we mostly want to disallow this
--- because the `where` disappears
+-- Bad - the `where` disappears
 someBinding mx = f <$> mx <*> mx where f x y = x * x + y * y
 
--- And this because the alignment is fiddly
+-- Bad - the alignment is fiddly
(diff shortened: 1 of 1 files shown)


jdreaver (approved):
Eh I think 1-spaced `where` is weird but I see the logic and I can get on board.

cbeav:
Agreed I thought this was weird too, but we're actually already using it in a surprising number of places in the codebase. I'm down for whatever as long as we're consistent, and I do like the vertical alignment above and below the `where` line.

pbrisbin:
Same here. I had a tough time acclimating to this, but I'm OK with it now. In my own 4-space world, this same style works out to 2-space/4-space instead of 1-space/2-space, so it's a little less weird.

Just to clarify, I'm not introducing this style here; it was already our preference. My intent here is to stop allowing inconsistencies and have the guide specify only one style.

cbeav (approved):

eborden:
Yeah, I also thought 1 space was odd at first, but I liked the alignment and context free love it brings.

Normal expression:
```
foo =
  bar
 where
  bar = "baz"
```

`do` notation:
```
foo = do
  bar
  bats
 where
  bar = putStrLn "baz"
```

eborden (approved):

5outh (approved):
I'm down with these changes. Good example coverage.

haskell-style.md
```diff
@@ -246,41 +246,36 @@ someBinding'' mx =
...
-   where -- PARSE ERROR - parser thinks this where is a part of the last expression
-    f x y = x * x + y * y
-
--- Instead, you have to do either this
-someBinding''' mx my = do
-  x <- mx
-  y <- my
-  f x y
+-- Good - use the same indentation for non-multi-line expressions to avoid
+-- having to change anythign if/when they grow more lines
```
cdparks: `anythign` -> `anything`

cdparks (approved):

PR #9, opened 2017-11-16
Add initial naming-conventions.md file

cdparks:
I'm adding this here instead of the wiki, but I'll probably link from there eventually.

Most of this is uncontroversial, but the thing we should discuss is the `Problems` section. From looking at our codebase, one might infer the following styles for enums:
1. Haskell - `TitleCase` and (rarely) `ALLCAPS`
2. Postgres - `snake_case` and `ALLCAPS`
3. JavaScript - `snake_case`, `TitleCase`, `camelCase`, `ALLCAPS`
4. YAML - `snake_case`, `hyphen-case`, `TitleCase`, `Title_Snake_Case` (?!)

Sometimes we translate `snake_case` enums into `TitleCase` or `camelCase` for the frontend, but we're definitely not doing it consistently.

Part of the problem is that we sometimes use the `To/FromJSON` instances to serialize enums to postgres, which means _there can only be one_ serialization method, and it usually fits the postgres convention. We also have some library functions `deriveJSONEnum[Using]` that seem to indicate at one point we did want to surface `snake_case` enums to the frontend.

**tldr;** it's more complicated than we thought. Plz discuss. Plz halp.

### DIFF
--- a/naming-conventions.md
+++ b/naming-conventions.md
@@ -42,9 +42,10 @@ foldr :: Foldable t => (a -> b -> b) -> b -> t a -> b
 traverse :: (Applicative f, Traversable t) => (a -> f b) -> t a -> f (t b)
 ```
 
-`m` is always always `Monad`
+`m` is almost always `Monad` or something monad-transformer related
 ```haskell
 when :: Monad m => Bool -> m () -> m ()
+liftIO :: MonadIO m => IO a -> m a
 ```
 
 `s` is for state and `e` is for environment
@@ -57,7 +58,7 @@ These are conventions only, but picking something unusual for a common typeclass
 
 ### Concrete Types
 
-Avoid using single-letter names or abbreviations for values with a concrete type, especially if that type is domain-specific.
+Avoid using single-letter names or abbreviations for values with a concrete type, especially if that type is domain-specific. To crib from the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#fundamentals), "Clarity is more important than brevity."
 
 For example, don't do this:
 ```haskell 
@@ -91,7 +92,7 @@ fetchCourseMemberships (Entity teacherId teacher) =
     pure memberships
 ```
 
-One abbreviation we seem to tolerate is `num` as in `numQuestionsAnswered`. There may be others.
+We tolerate a few abbreviations as part of identifiers (e.g. the `num` portion of `numQuestionsAnswered`), but there doesn't seem to be any broad consensnus about this.
 
 ## JavaScript
 
--- a/naming-conventions.md
+++ b/naming-conventions.md
@@ -92,7 +92,7 @@ fetchCourseMemberships (Entity teacherId teacher) =
     pure memberships
 ```
 
-We tolerate a few abbreviations as part of identifiers (e.g. the `num` portion of `numQuestionsAnswered`), but there doesn't seem to be any broad consensnus about this.
+We tolerate a few abbreviations as components of identifiers (e.g. the `num` portion of `numQuestionsAnswered`), but there doesn't seem to be any broad consensus about this.
 
 ## JavaScript
 
--- a/naming-conventions.md
+++ b/naming-conventions.md
@@ -2,23 +2,32 @@
 
 ## Postgres
 
-Postgres uses `snake_case`. Tables should be plural (e.g. `teachers`, `students`). Text-based enums are currently also `snake_case` (except domain abbreviations and math system which are `ALLCAPS`), though this makes interoperating with Haskell and JavaScript non-obvious.
+Postgres uses `snake_case`. Tables should be plural (e.g. `teachers`,
+`students`). Text-based enums are currently also `snake_case` (except domain
+abbreviations and math system which are `ALLCAPS`), though this makes
+interoperating with Haskell and JavaScript non-obvious.
 
 ## Haskell
 
-Haskell uses `camelCase` for identifiers and `TitleCase` for types and constructors. The latter is enforced by the language. Names that are being ignored should either use the wildcard `_` or specify what is being ignored (e.g. `_teacher`, `_standard`).
+Haskell uses `camelCase` for identifiers and `TitleCase` for types and
+constructors. The latter is enforced by the language. Names that are being
+ignored should either use the wildcard `_` or specify what is being ignored
+(e.g. `_teacher`, `_standard`).
 
 ### Polymorphic Types
 
-Prefer short, single-letter names for polymorphic type variables and arguments. You'll notice a number of conventions with type variables.
+Prefer short, single-letter names for polymorphic type variables and arguments.
+You'll notice a number of conventions with type variables.
 
 Unconstrained type variables often start with `a` and continue alphabetically
+
 ```haskell
 id :: a -> a
 const :: a -> b -> a
 ```
 
 Polymorphic lists are often "pluralized" by appending `s`
+
 ```haskell
 filter :: (a -> Bool) -> [a] -> [a]
 filter _ [] = []
@@ -30,6 +39,7 @@ filter predicate (a:as)
 ```
 
 `f` is for `Functor` and sometimes `Applicative` and `Alternative`
+
 ```haskell
 fmap :: Functor f => (a -> b) -> f a -> f b
 liftA2 :: Applicative f => (a -> b -> c) -> f a -> f b -> f c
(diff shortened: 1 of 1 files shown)


stackptr:
Could a naming convention for endpoint responses also be included? I.e., if JSON is returned from an endpoint it should follow the JS convention of camelCase for identifiers (fields). Otherwise frontend is forced to use custom parsing to enforce this

cdparks:
@stackptr yes, JSON fields from the backend are, as far as I can tell, universally `camelCase` with a few (very old) exceptions from when the API had just been ported from clojure. That is, we're consistent on fields, not on enums.

stackptr:
You're absolutely right, I was thinking of the enums (that is, the enums as they are represented in the JSON values).

naming-conventions.md
```diff
@@ -0,0 +1,106 @@
...
+guard :: Alternative f => Bool -> f ()
+```
+
+`t` might be `Traversable` or `Foldable`
+```haskell
+foldr :: Foldable t => (a -> b -> b) -> b -> t a -> b
+traverse :: (Applicative f, Traversable t) => (a -> f b) -> t a -> f (t b)
+```
+
+`m` is always always `Monad`
```
5outh: Or some variation on it, like `MonadIO`

naming-conventions.md
```diff
@@ -0,0 +1,106 @@
...
+    where_ $
+      case teacherSchoolId teacher of
+        Just schoolId ->
+          teachers ^. TeacherSchoolId ==. val (Just schoolId)
+        Nothing ->
+          teachers ^. TeacherId ==. val teacherId
+    pure memberships
+```
+
+One abbreviation we seem to tolerate is `num` as in `numQuestionsAnswered`. There may be others.
```
5outh: Eh, I don't know if we should explicitly say this because naming it `numQuestionsAnswered` is better imo

naming-conventions.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Problems
+
+Serialization of string-enums to JSON and postgres is all over the place. We have examples of `snake_case`, `camelCase`, `TitleCase`, and `ALLCAPS`.
+
+If we have DB-level enums that are not shared with the frontend, it makes sense to use the postgres convention of `snake_case`. However, if we're planning to pass enums to the frontend, then we should use whatever the JavaScript convention is. JavaScript is also using pretty much everything, so it's hard to say what we should be using. I (Parks) usually use `TitleCase` for JavaScript-only enums, but only because that mimics Haskell enums.
+
+Haskell pretty much enforces a style at the language-level by requiring that the first character of each constructor be capitalized. We're using `TitleCase` everywhere except for `MathSystem` and `DomainAbbreviation` which are `ALLCAPS`.
```
5outh: Is there a rule we should be following to determine if a new data type should be `TitleCase` or `ALLCAPS`?

naming-conventions.md
```diff
@@ -0,0 +1,106 @@
...
+
+## JavaScript
+
+JavaScript uses `camelCase` for identifiers and `TitleCase` for classes and React components. For flow types, we generally append a `T` to the end of the type, e.g. `SkillT`. Names that are being ignored should be prefixed with an underscore (e.g. `_teacher`, `_standard`). It's a common mistake (especially for Haskellers) to use a lone `_` and accidentally overwrite underscore in that scope.
+
+## Problems
+
+Serialization of string-enums to JSON and postgres is all over the place. We have examples of `snake_case`, `camelCase`, `TitleCase`, and `ALLCAPS`.
+
+If we have DB-level enums that are not shared with the frontend, it makes sense to use the postgres convention of `snake_case`. However, if we're planning to pass enums to the frontend, then we should use whatever the JavaScript convention is. JavaScript is also using pretty much everything, so it's hard to say what we should be using. I (Parks) usually use `TitleCase` for JavaScript-only enums, but only because that mimics Haskell enums.
```
5outh: If we have some enumerated column in a database entity, and spit it back to the frontend in the traditional way, `snake_case` will bleed into the frontend. This has a good chance of happening frequently, and without much thought.

I think it would be good to settle on a convention here. Couple of things to consider:

- When we pass an enum value back to the frontend, can we pre-process it into a different casing format? Can this be done by modifying typeclass instances?
- Can we do the same thing on the frontend by immediately converting `snake_case` values to `TitleCase`?
- Is it ever acceptable to use `camelCase` for enums?
- How much work would it take to:
  - Modify all database-level enums to be `TitleCase`?
  - Modify the API/Frontend bridge enums to be `snake_case`?

It would definitely be nice to settle on something here. I think at the very least we should aim to outlaw some naming schemes in specific contexts.

naming-conventions.md
eborden: "tolerate" might be the issue here. Instead maybe we should turn this into a list of accepted abbreviations.

naming-conventions.md
eborden: The one problem here is using `JSON` encoding for database values that are not actually `JSON`. We should really stop doing that. If something is indead `JSONB` data then it should be stored in accorance with our `JSON` style guide. Otherwise we shouldn't leverage encoding functions that use `To/FromJSON`.

naming-conventions.md
cdparks: Sorry, this was unclear. I wasn't saying `num` by itself is tolerated (I don't think I've ever seen that), just that it's tolerated _in identifiers_

naming-conventions.md
cdparks: The only things that I've seen that are `ALLCAPS` are initialisms, e.g. `MD` (a domain abbreviation) or `CCSS` (Common Core State Standards). I've seen `ELA` spelled `ELA` and `Ela` though, and we pretty much always say `schoolId` not `schoolID`.

eborden: `ALLCAPS` is also an older JavaScript style for constants. This might be the case in some places.

naming-conventions.md
joris974: I will be totally in converting as soon as receive information in the frontend to the format the frontend should use (camelCase). But it turn out hard to do because content we receive directly from JSON.
- Also the fancy api path route JS file is using snake case

naming-conventions.md
joris974: YEP! We use `ALLCAPS` for some constant definition but totally not attached to it. Will be happy to change to standardize

stackptr: I use and have seen the style `ALL_CAPS` for constants in JS. Is there an alternative convention we could use?

cdparks: I think `ALL_CAPS` for constants is fine, e.g.
```jsx
const MAX_FP_SESSIONS = 2
```

I was saying I've seen `ALLCAPS` for enum values, e.g.
```jsx
type MathSystemT
  = 'CCSS'
  | 'TEKS'
  ...
const MathSystems =
  { CCSS: ('CCSS': MathSystemT)
  , TEKS: ('TEKS': MathSystemT)
  ...
  }
```

cdparks: I don't have a good heuristic for formatting short initialisms. I've seen both `HTTP` and `Http` on the backend for example.

naming-conventions.md
5outh: Oooooh, I see. That was a little unclear haha

naming-conventions.md
5outh: In general I prefer `Http` (and `Ela` over `ELA` for example) in identifiers, because they'll always start with a lowercase letter and that's easier. For example:

`elaWritingAssignment` <-> `makeElaWritingAssignment` seems more consistent than:
`elaWritingAssignment` <-> `makeELAWritingAssignment`

But on it's own I think I'd prefer `ELA`. Does that seem like a reasonable rule to set up? If abbreviations are on their own => capitalize, if as a part of an identifier => camelCase?

eborden (changes requested):
Are we planning to update the "Problems" section with what you should do?

cdparks:
@eborden, yeah. I'd like to, but I'm not sure what the consensus is. Maybe I'll put my own preferences and find out who disagrees 😛

eborden:
I'm always in support of that path. Nothing gets people to disagree with you like writing something down :grinning:

cdparks:
Replaced `Problems` section with draft `Interoperation` section. The assertions there do not necessarily match the current reality of the codebase. Additionally, whatever we decide, we should make canonical TH generators (or `genericToWhatever` functions) and delete everything else.

In fact, I would argue that anything that has to deviate from the preferred style should use a hand-written instance to call it out as being intentional

naming-conventions.md
```diff
@@ -0,0 +1,188 @@
...
+
+We tolerate a few abbreviations as components of identifiers (e.g. the `num` portion of `numQuestionsAnswered`), but there doesn't seem to be any broad consensus about this.
+
+## JavaScript
+
+JavaScript uses `camelCase` for identifiers and `TitleCase` for classes and React components. For flow types, we generally append a `T` to the end of the type, e.g. `SkillT`. Names that are being ignored should be prefixed with an underscore (e.g. `_teacher`, `_standard`). It's a common mistake (especially for Haskellers) to use a lone `_` and accidentally overwrite underscore in that scope.
+
+## {DRAFT} Interoperation Between Languages
+
+JSON keys should always be `camelCase`. String enums that cross language barriers should be `snake_case`, since most enums that cross language barriers are designed for `postgres` first. Exceptions can be made for short initialisms which may be `ALLCAPS` (e.g. `TEKS: MathSystemT`). Short initialisms that are part of other identifiers should use `snake_case` (e.g. `rti_coordinator: TeacherRoleT`).
```
eborden: Not sure I agree that "string enums that cross language barriers should be `snake_case`." What makes this necessary? We have separate encoding/decoding functions for postgres/json. We should be able to support each environment's style.

eborden (changes requested):

naming-conventions.md
cdparks: We have separate encoding/decoding functions for postgres/json, but we also have JSON that's stored in the DB with embedded enums (some of which are already `snake_case`).

pbrisbin (approved):
This is all very well stated and I have no issues with the conventions as outlined. Nice job.

Things I'd be interested in investigating/discussing:

1. How much existing code does this invalidate, and do we have any way of itemizing it and/or plans to address it?

1. Are there any ideas for enforcing (or at least programatically encouraging) this going forward, for example could we declare our derivation functions such that the easy thing is also the correct thing?

1. Would something like this always Just Work?

   ```hs
   preferredEnumify t
     | T.all isUpper t = t
     | otherwise = snakeCaseify $ unCapitalize t
   ```

pbrisbin (commented):
I reviewed this in "rich diff", so I have a few follow-up Markdown nits now that I saw the source:

1. Hard-wrap lines at 80 columns (editors can be configured to handle this)
1. Surround block elements with a line of whitespace; in this case, you're missing an empty line before each of your code blocks.

These are my own (but very common) styles for markdown, so feel free to ignore if you don't prefer them -- we have no accepted style-guide on markdown yet.

naming-conventions.md
eborden: I guess this is the difference between this doc being inspirational or consistent with existing style?

cdparks:
@pbrisbin 
1. I'm not sure. I would want to add a tech debt task to audit our codebase after we have some idea about what we want.
2. Yes, we would definitely want to declare canonical derivation functions (and probably delete anything we're not using)
3. Probably

Also re: markdown, I created and edited this online using the GitHub editor :smile: 

Fixed locally in vim of course.

naming-conventions.md
cdparks: Yeah, I don't think there is a way to be consistent with our existing style :stuck_out_tongue:

eborden (approved):

PR #10, opened 2017-11-29
Initial Shell style guide

pbrisbin:
I have more to add, but I thought I'd get the discussion started while I put
this down for a bit and come back later.

### DIFF
--- a/shell-style.md
+++ b/shell-style.md
@@ -40,7 +40,7 @@
    scripts to run and selecting an explicit shell will ensure consistent
    behavior.
 
-1. Use `#!/usr/bin/env bash`
+1. Use `#!/usr/bin/env bash`, not `#!/bin/bash`
 
    This is specifically for OSX folks, where `/bin/bash` is severely out of
    date. A newer version, installed via Homebrew, may appear first in `$PATH`
--- a/shell-style.md
+++ b/shell-style.md
@@ -75,8 +75,9 @@ if you use `[[` because you want regex or glob matching in one place, don't use
 
 2. Avoid `set -x`
 
-   Only use this if your script has no function definitions, control flow
-   structures, or conditional expressions -- otherwise it gets too noisy.
+   This option outputs every statement before it's executed. Only use this if
+   your script has no function definitions, control flow structures, or
+   conditional expressions; otherwise it is overly noisy.
 
    ```sh
    # Bad
--- a/shell-style.md
+++ b/shell-style.md
@@ -1,0 +1,342 @@
+# Shell Style
+
+## Global Rules
+
+1. All scripts should pass [Shellcheck][]
+
+   [shellcheck]: https://www.shellcheck.net/
+
+   Add explicit annotations as needed. When adding new scripts confirm, that our
+   CI machinery is configured to run on your files.
+
+1. All scripts end in a `.sh` extension
+
+   This is typically an anti-pattern, but we currently rely on it to locate
+   shell scripts within the source tree for the purposes of Shellcheck on CI.
+
+1. Always use double-quotes, except where interpolation must be prevented
+
+   NOTE: It's acceptable to use single-quotes as a matter of course for script
+   arguments to other interpreters like `sed`, `awk`, and `psql` even if there
+   would be no unintentional interpolations currently.
+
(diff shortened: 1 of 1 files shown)


shell-style.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Interpreter
+
+1. Prefer Bash
+
+   We write all of our scripts as Bash. It's available everywhere we need
+   scripts to run and selecting an explicit shell will ensure consistent
+   behavior.
+
+1. Use `#!/usr/bin/env bash`
```
5outh: Maybe include

> instead of the normal shebang line `#! /bin/bash`

for clarity?

shell-style.md
```diff
@@ -0,0 +1,342 @@
...
+   `-u` will halt the script if an unset variable is used.
+
+   This avoids bugs like:
+
+   ```sh
+   dir=/tmp/foo
+   rm -rf "$dirr/bar" # typo made this rm -rf /bar
+   ```
+
+2. Avoid `set -x`
```
5outh, line 76: Maybe include what `-x` does and when it should be considered

shell-style.md
```diff
@@ -0,0 +1,342 @@
...
+   ```sh
+   # Bad, suprisingly
+   foo="barbaz"
+   foo="$bar$baz"
+
+   # Good
+   foo=barbaz
+   foo=$bar$baz
+
+   # But required fo
```
5outh: typo `fo -> for`

shell-style.md
```diff
@@ -0,0 +1,342 @@
...
+   foo="bar baz"
+   foo="$bar $baz"
+   ```
+
+## Best Practices
+
+1. Scripts should have a header comment, formatted like so:
```
5outh, line 162: :+1:

eborden:
I'm going to give this a review, but I have very few opinions on how `sh` should be written. I'll likely be happy to accept any form of standardization.
It would be good to note that we use https://www.shellcheck.net/ currently.

5outh (approved):
Looks like a really nice guide! I had a couple of suggestions, but great stuff overall.

I'd also be interested in a section "When *to* Use Shell" in the same vein as the "When Not to Use Shell" - when we have all of these alternatives, what is the right reason to reach for this tool? Just something to consider.

Awesome stuff! :+1:

shell-style.md
```diff
@@ -0,0 +1,347 @@
...
+   fi
+
+   while; do
+   done
+
+   for x; do
+   done
+   ```
+
+1. Avoid redundant braces:
```
jdreaver, line 131: I'm not sure I agree with this. I think the point of a lot of good bash style is to avoid the danger of innocuous modifications breaking something in the future. For example, even if you know a variable has not spaces, you quote it when used in case someone adds a space in the future.

Someone could very easily change

```sh
touch "foo/$bar/baz"
```

into 

```sh
touch "foo/$bar_backup/bat
```

Obviously we should be testing changes we make to scripts, but for particularly noisy strings with a variable (especially file paths!) sometimes I like to add the brackets to highlight the interpolation.

jdreaver (approved):
This is awesome, thanks!

eborden (approved):
This is a very well written guide. Thanks for doing it!

shell-style.md
```diff
@@ -0,0 +1,347 @@
...
+   touch "foo/$bar/baz"
+
+   # But required here
+   touch "foo/${bar}_backup/bat
+   ```
+
+1. Avoid redundant quoting on assignment:
+
+   ```sh
+   # Bad, suprisingly
```
cdparks, line 147: What kind of bad things can happen here?

pbrisbin: "Bad" is misleading (it's just the language we use for all style-guide counter-examples). It's just unnecessary.

EDIT: and I said "surprisingly" because generally shells want you to quote absolutely everything and here it's not necessary.

cdparks: Ah, ok, yeah, I've always learned quote everything forever and ever amen.

pbrisbin:
Even with an open thread, there are a lot of green checkmarks so I'm going to merge now rather than let this go stale. Happy to follow up with more changes and additions in subsequent PRs.

PR #15, opened 2017-12-14
Canonicalize operator-first style

### LINKED ISSUE #11: Operator-first / Applicative
There are a few styles for `Applicative` expressions

My personal preference:

```hs
parseJSON = withObject "Foo" $ \o -> Foo
  <$> o .: "this"
  <*> o .: "that"
```

Following a broader (but not documented) "operator-last" guide:

```hs
parseJSON = withObject "Foo" $ \o -> Foo <$>
  o .: "this" <*>
  o .: "that"
```

And each of these comes in a "newlined" flavor:

```hs
parseJSON = withObject "Foo" $ \o ->
  Foo
  <$> o .: "this"
  <*> o .: "that"

parseJSON = withObject "Foo" $ \o ->
  Foo <$>
  o .: "this" <*>
  o .: "that"
```

We of course have a mix across our code-bases now; I'd like to standardize that one way or another.

---

I'm not particularly against operator-last as a general guide, however for expressions such as `Applicative`, I strongly prefer operator-first. And I believe much of the team agrees.

There are other expressions in the same category, which I would call "structural" expressions. For example, a lens chain like:

```hs
foo
  <&> bar .~ baz
  <.> bat .~ quiix
```

I call these structural because they look like records or lists. And I think a lot of the team (even if operator-last in general) likes operator-first here because it mirrors the comma-first style of records and lists.

This is in contrast to "functional" expressions, which is where I think folks generally prefer operator-last style:

```hs
foo . bar . baz $
  bat quix

-- vs
foo . bar . baz
  $ bat quix
```

---

So, I would like to raise some questions to the team:

1. Would folks be opposed to prefer-operator-first as our overall guide?
1. If so, would folks be opposed to prefer-operator-first for structural expressions?
1. If so, would folks be opposed to prefer-operator-first for `Applicative` expressions?
1. If so, would folks be opposed to making prefer-operator-**last** explicit in these guides?
1. If so, that means we're OK letting folks do whatever they personally want, are you sure?
1. If we're making a new guide, are we enforcing the newlined constructor or not?

pbrisbin:
There are some changes to surrounding, related content, which are in isolated
commits. Let me know if any should be reverted.

Closes #11.

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -114,7 +114,9 @@ Signatures should lead with arrows
 
 ## Alignment
 
-You should never need to align to preceding text. That is, your aligment point should always be 3rd, 5th, 7th etc. column (because we use two spaces of indenting).
+You should never need to align to preceding text. That is, your alignment point
+should always be 3rd, 5th, 7th etc. column (because we use two spaces of
+indenting).
 
 e.g. Instead of
 
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -130,20 +130,34 @@ instance FromJSON (GameSession Never MathStandardAssignmentId) where
                 <*> o .: "sub-sub-standard-perc"
                 <*> o .: "coins-gained"
                 <*> pure Never
- ```
-which will need cause a noisy diff if we rename `GameSession` (each following line also needs to be indented)
-```haskell
-instance FromJSON (GameSession Never MathStandardAssignmentId) where
-  parseJSON = withObject "cannot parse GameSession" $ \o ->
-    GameSession2 <$> o .: "answers"
-                 <*> o .: "domain-id"
-                 <*> o .: "current-standard"
-                 <*> o .: "sub-standard-perc"
-                 <*> o .: "sub-sub-standard-perc"
-                 <*> o .: "coins-gained"
-                 <*> pure Never
 ```
-do
+
+which will cause a noisy diff if we rename `GameSession` (each following line
+also needs to be indented):
+
+```diff
+
+ instance FromJSON (GameSession Never MathStandardAssignmentId) where
+   parseJSON = withObject "cannot parse GameSession" $ \o ->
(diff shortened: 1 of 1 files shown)


haskell-style.md
```diff
@@ -156,6 +172,88 @@ instance FromJSON (GameSession Never MathStandardAssignmentId) where
         <*> pure Never
 ```
 
+```diff
+ instance FromJSON (GameSession Never MathStandardAssignmentId) where
+   parseJSON =
+     withObject "cannot parse GameSession" $ \o ->
+-      GameSession
++      GameSession2
```
eborden, line 180: Seems like we should figure out a consensus on this style point. I'm totally down to get rid of this and prefer no newline.

eborden (changes requested):

haskell-style.md
pbrisbin: Yeah, me too. I think it's a broader discussion though, to be had separately from operator-first. For example, I know @cdparks 

```hs
uses =
  newlines
    $ like mad
```

And I'm happy to let him, while I don't prefer it myself.

So, I'd be happy to not explicitly mention it here (so revert 6d64f2d, basically) and pick up that discussion separately.

I'm also down to leave this ambiguous for now, and bank on Brittany making it [moo](https://assets-auto.rbl.ms/e918f7c316ce8442edba7a0c9e58c62fe0d3b4996eb6b1e04d3f2c6b0481289c).

eborden: :laughing: moo :rofl:

pbrisbin:
New 100% rebased! Now with 0% newline-controversy!

eborden (approved):

haskell-style.md
cdparks:
```haskell
this =
  is
    a
    fair
    characterization
```

PR #16, opened 2018-04-06
Update default-extensions rule

### LINKED ISSUE #13: default-extensions rule
The current rule is we do not use `default-extensions`; extensions should always be specified in the module that uses them.

We should revisit:

1. I believe people didn't realize this was coming in when it was introduced
1. Some cases such as `NoImplicitPrelude` really should be in defaults

This issue serves as additional awareness, to resolve (1) even if we change nothing.

To address (2), I would propose we change the guide to specify an explicit list of extensions that are allowed in `default-extensions` and enforce that all others are enabled by pragma. I would begin concervatively with `[NoImplicitPrelude]` being the safe-list, but welcome suggestions from others.

pbrisbin:
Closes #13.

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -588,12 +588,29 @@ module Driver
 
 ## Declaring Extensions
 
-- Declare extensions in the module(s) that use them, i.e. don't use
-  `default-extensions`
+- All Haskell packages MUST use the following `default-extensions`:
 
-  This ensures a) the source code works un-changed with other Haskell tooling
-  (e.g. `ghci`, `doctest`, etc) without further configuration and b) active
-  extensions are visible in the modules they impact.
+  ```yaml
+  default-extensions:
+    - BangPatterns
+    - DeriveGeneric
+    - LambdaCase
+    - NoImplicitPrelude
+    - NoMonomorphismRestriction
+    - OverloadedStrings
+    - RankNTypes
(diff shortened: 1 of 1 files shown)


haskell-style.md
```diff
@@ -588,12 +588,29 @@ module Driver
...
+    - NoMonomorphismRestriction
+    - OverloadedStrings
+    - RankNTypes
+    - RecordWildCards
+    - ScopedTypeVariables
+    - TypeApplications
+    - TypeFamilies
+  ```
+
+  **NOTE**: `NoImplicitPrelude` may be omitted in packages using the normal,
```
eborden, line 608: Does this jive with tooling?

eborden: I think it does, since it doesn't have any syntactic effects.

eborden (approved):

haskell-style.md
pbrisbin: I think you're right in that if a syntax-only tool has `NoImplicitPrelude` in its own extensions configuration, but a package does not (and so doesn't `import Prelude`) it still works because it's syntax-only. So I think we're good.

I was worried for a second about DocTest, but since you still define those extensions in the executable in the package itself, you can just make it consistent either way. Similar story for a package-local `.ghci`.

5outh (approved):

PR #17, opened 2018-05-10
Derive all the things

eborden:
Extend the list of default extensions to include the full suite of deriving extensions. We want making types to be cheap and for types to be usable they often need to derive helpful typeclasses. By turning on the full suite of derivation extensions we make deriving cheaper and thus new types cheaper.

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -593,7 +593,14 @@ module Driver
   ```yaml
   default-extensions:
     - BangPatterns
+    - DeriveAnyClass
+    - DeriveFoldable
+    - DeriveFunctor
     - DeriveGeneric
+    - DeriveLift
+    - DeriveTraversable
+    - DerivingStrategies
+    - GeneralizedNewtypeDeriving
     - LambdaCase
     - NoImplicitPrelude
     - NoMonomorphismRestriction
@@ -601,6 +608,7 @@ module Driver
     - RankNTypes
     - RecordWildCards
     - ScopedTypeVariables
+    - StandaloneDeriving
     - TypeApplications
     - TypeFamilies
   ```


cdparks (approved):

5outh (approved):
Niiiice!

haskell-style.md
```diff
@@ -593,14 +593,22 @@ module Driver
...
+    - DerivingStrategies
+    - GeneralizedNewtypeDeriving
     - LambdaCase
     - NoImplicitPrelude
     - NoMonomorphismRestriction
     - OverloadedStrings
     - RankNTypes
     - RecordWildCards
     - ScopedTypeVariables
+    - StandaloneDeriving
```
pbrisbin, line 611: I'm not sure about this one. I don't think it totally aligns with the stated motives of this PR, and (IMO) it promotes Orphans. Would you consider merging without it?

cdparks: I think we would catch orphans in code review, plus a submitter would have to manually suppress the warning. The other use case for this is deriving instances for GADTs which don't support stock deriving. 

The parser catches you doing it when it's turned off so I don't think it's adding any overhead

cdparks (approved):

haskell-style.md
eborden: Yeah, standalone deriving isn't about orphans, it is just about being able to enumerate extra constraints that GHC can't infer in normal deriving.

pbrisbin (approved):
I don't really follow the `StandaloneDeriving` stuff, but I concede.

eborden:
@pbrisbin here is an example where GHC needs extra constraints to do its thing:
```
deriving instance Show (IdpId idp) => Show (FailedIdpSyncItems idp)
deriving instance Eq (IdpId idp) => Eq (FailedIdpSyncItems idp)
```

PR #21, opened 2018-06-15
Describe preferred Haddock formatting style

pbrisbin:


haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+    -- | Foo's bar
+    , fooBar :: Bar
+    }
+  -- ^ What even is a Foo?
+  ```
+
+  Good
+
+  ```hs
+  -- | What even is a Foo?
```
cdparks: Indeed

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+-- of the fact that they are this thing
+
+-- | Does a thing. Is partial because of random reason
+```
+
+Good
+
+```hs
+-- | Represents a value that may or may not be present
+{- This type represents a value... -}
```
cdparks: The `{- ... -}` comments look like they're meant to represent the comment body. Is this supposed to be showing how the summary completes the "This <thing> ..." sentence?

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+    , deleteFoo
+    )
+  ```
+
+  Good
+
+  ```hs
+  module Foo
+    (
+    -- * Foo
```
cdparks, line 933: I think some guidance on sections would be useful. I've seen bare nouns as well as sentences here in our codebase

haskell-style.md
pbrisbin: > Is this supposed to be showing how the summary completes...

That was the intention, I can remove it if it's confusing.

haskell-style.md
pbrisbin: I'm open to guidance or leaving it un(der)-specified. What do you propose?

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+    , fooBar :: Bar
+    }
+  -- ^ What even is a Foo?
+  ```
+
+  Good
+
+  ```hs
+  -- | What even is a Foo?
+  data Foo = Foo
```
jdreaver: We should probably follow the convention for records in the rest of this guide :smile: 

```haskell
data Foo
  = Foo
  { ...
```

haskell-style.md
eborden: A convenient mispelling of `fu` to avoid detection when coupled with `bar`?

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+    { fooFoo :: Foo
+    -- ^ Foo's foo
```
jdreaver: Do we want to indent these comments by 2, so they aren't at a lower indentation than the thing they are describing?

```
data Foo
  = Foo
  { fooFoo :: Foo
    -- ^ Foo's foo
  ...
```

I'm fine either way on this, just throwing it out there.

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
   ```
+
+## Haddocks
+
+Haskell modules should be commented with valid Haddock documentation. We are not
+yet requiring a certain level of coverage, but it is strongly encouraged.
+
+### General guides
+
+- Use proper Haddock [markup][]
```
eborden, line 671: Is this meant to be a link to https://www.haskell.org/haddock/doc/html/index.html

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+
+  **NOTE**: do not align trailing documentation in context-sensitive ways.
+
+  Bad
+
+  ```hs
+  -- | What even is a Foo?
+  data Foo = Foo
+    { fooFoos :: [Foo] -- ^ Foo's foos
+    , fooBar :: Bar    -- ^ Foo's bar
```
jdreaver: I would add a really long comment here too for the Bad case

```haskell
  , fooBaz :: Baz -- ^ This is a really long comment and we should make sure we
                  -- don't align this in context-sensitive ways
```

Up to you

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+
+- Consider adding section documentation
+
+  **NOTE**: Summary/Body rules apply!
+
+  ```hs
+  module Foo
+    (
+    -- * Foo
+    -- | Operates on Foos
```
jdreaver, line 953: Same comment I had about indenting 2 spaces in records. Having the comment start 2 spaces less indented than the thing we are commenting feels a bit funny.

haskell-style.md
eborden: Yes please, much confuse.

jdreaver (commented):
Very nice!

haskell-style.md
```diff
@@ -660,3 +660,329 @@ module Driver
...
+  -- 2. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum
+  --    dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
+  --    proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+  --
+  theFunction
+  ```
+
+### Module organization
+
+- Organize your exports by logical groups or progressive disclosure and use
```
eborden: Any links for a description of progressive disclosure?

haskell-style.md
eborden: Oh wait, that is defined below. How have I never seen this link style in markdown?

eborden (changes requested):

haskell-style.md
jdreaver: haha, I had the exact same train of thought you have here.

haskell-style.md
eborden: It seems @pbrisbin is following the `brittany` style here (rather the proposed style, since that branch is still in dev).

jdreaver: Ah makes sense. I wonder if we should be consistent for now until we use `brittany`? I'm down to move to this eventually if that's what `brittany` does.

haskell-style.md
cdparks: For some packages in the wild, I've seen `Data Types`, `Functions`, which doesn't seem useful to me. I've also seen `Tutorial`, `Exports`, `<whatever the module's actually about>`, `Internal`. I dunno, I'm fine to let people improvise here until we have a better idea.

haskell-style.md
pbrisbin: Doh, A little _pbrisbin_ style slipped in.

haskell-style.md
pbrisbin: I've never seen that done that way, so I'm generally against sprouting our own convention here that differs with what you see in the wild. Then again, I haven't done an exhaustive search (or any really) so maybe that is a common way to do it?

haskell-style.md
pbrisbin: Oh yeah! I wanted to mention that the usual Summary/Body guides apply to trailing comments. And that trailing comments can only be on the same line if they're short enough.

So:

```hs
, fooBaz :: Baz
-- ^ The Baz of this foo
--
-- And  this is what you do when it gets really long and needs a Body.
--
```

If I don't get immediate push back on this comment, I'll add that to the PR.

haskell-style.md
pbrisbin: > Having the comment start 2 spaces less indented than the thing we are commenting feels a bit funny.

That's just an awkward thing that happens when documenting the first export in a module list (and it doesn't happen with attributes or arguments because they are documented trailing-style). I've gotten used to it, and I wouldn't want to add an indentation to the Guide overall because of this first element quirk.

That's the other perspective anyway, I'd love to hear other voices from the team.

cdparks: Yeah, I think it's fine. It's aligning the comments with the commas, there's just no leading comma. You could imagine trying to do:
```haskell
module Foo
  -- * Foo
  -- | Operates on Foos
  ( getFoo
  , deleteFoo

  -- * Bar
  -- | Operates on Bars
  , updateBar
  ...
  )
```

but Haddock rejects this

haskell-style.md
jdreaver: Never? I thought this was by far the most common

* `lens`: http://hackage.haskell.org/package/lens-4.16.1/docs/src/Control-Lens-TH.html (gotta ctrl + F for `-- ^`
* `aeson`: http://hackage.haskell.org/package/aeson-1.4.0.0/docs/src/Data.Aeson.Types.Internal.html#Options
* `shake` actually goes further than aligned https://hackage.haskell.org/package/shake-0.16.4/docs/src/Development.Shake.Internal.Options.html#ShakeOptions
* `time` is using the `-- |` style, but still aligning https://hackage.haskell.org/package/time-1.9.1/docs/src/Data-Time-LocalTime-Internal-TimeZone.html#TimeZone

I wouldn't call it "sprouting our own convention" at all.

haskell-style.md
jdreaver: Ah I wasn't talking about the first comma missing. I was talking about not aligning to the comma, but the identifier. In fact this style doesn't suffer from the awkwardness of the first comma missing

```haskell
module Foo
  (
    -- * Foo
    -- | Operates on Foos
    getFoo
  , deleteFo

    -- * Bar
    -- | Operates on Bars
  , updateBar
  ...
  )

Comment (cdparks) [reply haskell-style.md:953]: Both styles look fine to me, no strong opinion

File: haskell-style.md (hunk 5 above)
Comment (pbrisbin) [reply haskell-style.md]: - Lens

  This is doing exactly what I'm proposing, aligning with `,` with a weird first element.

   ```hs
   module Control.Lens.TH
    (
    -- * Constructing Lenses Automatically
    -- ** Lenses for data fields
      makeLenses, makeLensesFor
    , makeClassy, makeClassyFor, makeClassy_
    , makeFields
    , makeFieldsNoPrefix
    -- ** Prisms
    , makePrisms
    , makeClassyPrisms
    -- ** Wrapped
    , makeWrapped
  ```

- Aeson is doing what you suggest, :ok_hand: 

- Shake is doing what you suggest, if you take a liberal view since it doesn't put a space after `,` and is not aligning to identifier

  ```hs
    {shakeFiles :: FilePath
          -- ^ Defaults to @.shake@. The directory used for storing Shake metadata files.
          --   All metadata files will be named @'shakeFiles'\/.shake./file-name/@, for some @/file-name/@.
          --   If the 'shakeFiles' directory does not exist it will be created.
    ,shakeThreads :: Int
          -- ^ Defaults to @1@. Maximum number of rules to run in parallel, similar to @make --jobs=/N/@.
          --   For many build systems, a number equal to or slightly less than the number of physical processors
          --   works well. Use @0@ to match the detected number of processors (when @0@, 'getShakeOptions' will
  ```

  So I'm not sure it should count.

- Time

  I claim disqualified because it's not leading comma. It is actually aligning to the left margin, which could also be considered what I'm suggesting.

  ```hs
  -- | A TimeZone is a whole number of minutes offset from UTC, together with a name and a \"just for summer\" flag.
  data TimeZone = TimeZone {
      -- | The number of minutes offset from UTC. Positive means local time will be later 
  in the day than UTC.
      timeZoneMinutes :: Int,
      -- | Is this time zone just persisting for the summer?
      timeZoneSummerOnly :: Bool,
      -- | The name of the zone, typically a three- or four-letter acronym.
      timeZoneName :: String
  } deriving (Eq,Ord,Data, Typeable)
  ```

So you've got 4 examples: one in my favor, one in yours, one disqualified and one that's iffy :joy:

Comment (jdreaver) [reply haskell-style.md]: This is the comment thread for records, not imports! Lens is doing what I want haha

File: haskell-style.md (hunk 3 above)
Comment (pbrisbin) [reply haskell-style.md:933]: > I dunno, I'm fine to let people improvise here until we have a better idea.

That's my take. It's unclear if these are Summary (which should then be declarative answers to "This section... {Summary}" or just Titles/Topics/Headings (which is why `Foo` or `Internal` works). Because of this, I was fine leaving this bit open-ended.

File: haskell-style.md (hunk 5 above)
Comment (pbrisbin) [reply haskell-style.md]: Are you proposing using a different style for exports vs records? I have a strong opinion that we can and should use a consistent guide for both elements, which is align to the `,`. This is simpler and simpler is better (IMHO).

Comment (jdreaver) [reply haskell-style.md]: No, I'm proposing aligning to the identifier on both, not the comma, since we aren't describing commas we are describing identifiers.

I'm fine with the style in this PR if I'm outvoted, I just think it looks bad and I was offering an alternative. I was also trying to dispel the idea that I was just "sprouting" this style out of nowhere.

**Comment by pbrisbin** [comment]
I think I've addressed all open questions except for the indentation question @jdreaver has raised. There was some Thread overlap :confused: so I'm trying to summarize here and continue the discussion at top-level:

   A: aligned to left-margin

   ```hs
   module Foo
     (
     -- * Foo
       Foo
     
     -- * Bar
     , Bar
     ) where

   data Foo
     = Foo
     { fooFoo :: Foo
     -- ^ Foo's Foo
     , fooBar :: Bar
     -- ^ Foo's Bar
     }

   fooTheFoo
     :: Foo
     -- ^ The Foo argument
     -> Boo
     -- ^ The Bar argument
   ```

   B: indented / aligned to identifier (depending on how you look at it)

   ```hs
   module Foo
     (
       -- * Foo
       Foo
     
       -- * Bar
     , Bar
     ) where

   data Foo
     = Foo
     { fooFoo :: Foo
       -- ^ Foo's Foo
     , fooBar :: Bar
       -- ^ Foo's Bar
     }

   fooTheFoo
     :: Foo
       -- ^ The Foo argument
     -> Boo
       -- ^ The Bar argument
   ```

I _think_ the options/preferences on the table are:

1. (A) for exports and attributes and arguments (me)
1. ~~(A) for exports and arguments, but (B) for attributes (David)~~
1. (B) for exports and attributes and arguments (David)

Is that an accurate summary of the question(s) at hand?

**Comment by jdreaver** [comment]
Nah I'm B all the way for all 3 of these things (exports, records, args), although like you mentioned in this guide if it is short I prefer just appending too

```haskell
data Foo
  = Foo
  { fooFoo :: Foo -- ^ Foo's Foo
  , fooBar :: Bar -- ^ Foo's Bar
  }

fooTheFoo
  :: Foo -- ^ The Foo argument
  -> Boo -- ^ The Bar argument
```

My reasoning is thus: we are describing an identifier, not a `,` or `::`, `->`. We should align to the identifier in question, not 2 spaces before it. Commas are separators for list items, not the beginning of list items.

pbrisbin:
I'll update that comment to correct your preference.

> My reasoning is thus: ... we should align to the identifier in question

But that reasoning breaks with function arguments, unless you're proposing 3-space indent there. And even if you go 3-space indent, your approach still raises more questions with functional function arguments:

```hs
fooTheFoos
  :: FooIsALongTypeSoLetsNotDoSameLine
    -- ^ This documentation is just indented, not identifier aligned

fooTheFoos
  :: FooIsALongTypeSoLetsNotDoSameLine
     -- ^ This is identifier aligned, but it's odd indention

fooTheFoos
  :: (Foo -> Bar -> Baz)
     -- ^ Is "(" the identifier?
```

We do "odd" indentation for `where`, so that may be fine, but I think the raised question about `(` is a point against.

I think we might just need a team-wide vote? You and I just disagree in preference here.

jdreaver:
Hmm, yeah function arguments are odd. I admit I don't think I've ever used `-- ^` on the line after a function argument anyway, just immediately after the argument on the same line.

pbrisbin:
Yeah, it's tricky. I think that's the core of my argument against your style:

"Identifier-aligned" is a form of context-sensitive, in that it is different indentations in different contexts, where the identifier is preceded by different numbers of non-identifier characters.

And we've established anti-context-sensitive as a sort of golden meta-rule.

If this weren't the case (if your guide applied cleanly to all contexts) I would've yielded by now. Your arguments are compelling otherwise. But I think I want a simple guide that we can apply the same way to exports, attributes, and arguments.

I think if you were just arguing for indenting it, I might yield. It'd only be coincidental that that results in identifier-aligned in some contexts (and only because we use 2-space indent). But you've already given away that that wasn't the proposal :)

jdreaver:
I'd still propose my style for exports and records though. Indeed it is just indenting two more spaces. I wonder how common next-line comments even are for arguments, we could ban them entirely.

pbrisbin:
@cdparks @eborden @5outh can you all please vote on this comment?

:+1: for the non-indented versions currently in the PR proper

:-1: for David's style: indenting for exports and attributes, but **not** for function arguments:

```hs
module Foo
  (
    -- * Foo
    Foo

    -- * Bar
  , Bar
  ) where

data Foo
  = Foo
  { fooFoos :: Foo
    -- ^ Foo's foos
  , fooBars :: Bar
    -- ^ Foo's Bar
  }

fooTheFoo
  :: Foo
  -- ^ The Foo argument
  -> Bar
  -- ^ The Bar argument
  -> IO ()
fooTheFoo
```

Please comment if you would consider banning next-line comments in function arguments to make David's style easier to describe and understand.

cdparks:
I use next-line argument docs all the time:

```haskell
-- | Post a new reading answer or retry for the specified assignment
postArticleReadingArtifactR
  :: forall assignment session answer retry artifact
  . ( AssignmentFamilyWithRetries assignment session answer retry
    , IsArticleReadingAssignment assignment
    , IsArticleReadingSession session
    , IsArticleReadingAnswer answer
    , IsArticleReadingRetry retry
    )
  => (assignment -> IsSelfAssigned)
  -- ^ How to determine whether assignment was teacher or student-assigned
  -> ([Entity retry] -> ResumableSessionAttempt)
  -- ^ How to determine which attempt this is
  -> HandleRunDB (ResumableSessionResponse (Entity artifact))
  -- ^ Handle exceptions from runDB
  -> InsertArtifact session answer retry artifact
  -- ^ Insert and validate new artifact
  -> AssignmentType
  -- ^ Assignment type for badges
  -> NonEmpty (Key assignment)
  -- ^ Assignment to post answer to. NB - only first assignment Id is used
  -> Handler (ResumableSessionResponse (Entity artifact))
```

Or in type synonyms:

```haskell
-- | Generic constructor for answers of different types
type MkAnswer session answer
  = Key session
  -- ^ Session Id
  -> ElaArticleReadingQuestionId
  -- ^ Question Id
  -> JSONB ReadingAnswerIndexes
  -- ^ Student answer choices
  -> Percentage Double
  -- ^ Correctness
  -> UTCTime
  -- ^ Created at timestamp
  -> answer
```

jdreaver:
@cdparks SGTM then, I rescind my comment about banning them :smile:

pbrisbin:
Worth noting: we can't appeal to WWBD (what would brittany do) here. I tested in their little paster thing and it just left the comments where they were in either style.

cdparks:
In general I'm :+1:  on "A: aligned to left-margin", though probably just because I'm used to it. I can see both styles as some kind of "context-sensitive" but aligning to identifier has more jumping around from line to line, so I prefer aligning to the margin

eborden (approved):

pbrisbin:
So, @cdparks voted in his own comment rather than on that comment, and @eborden Approved the PR overall, which I guess also counts as a vote for the un-indented form?

I _think_ this means we have a majority. @5outh can you weigh in?

5outh:
I'm for the non-indented versions; I write comments like that constantly!

5outh (approved):

pbrisbin:
Thank you all for spending your time settling this. Our documentation will be better for it :smile:

PR #24, opened 2018-07-12
Add Foreign Keys to naming conventions

cdparks:
I've noticed we have this problem in a few of our ELA-related entities, and it makes changing them difficult. @pbrisbin and I already chatted a bit about this over Slack. Let me know if y'all have thoughts.

### DIFF
--- a/naming-conventions.md
+++ b/naming-conventions.md
@@ -7,6 +7,53 @@ Postgres uses `snake_case`. Tables should be plural (e.g. `teachers`,
 abbreviations and math system which are `ALLCAPS`), though this makes
 interoperating with Haskell and JavaScript non-obvious.
 
+### Foreign Keys
+
+If a table has a foreign key to another table, remove the _product_ and
+_subject_ prefix from the foreign key column's name unless doing so would
+introduce ambiguity.
+
+Don't do this:
+
+```sql
+CREATE TABLE ela_adaptive_skill_practice_paragraphs (
+  id integer PRIMARY KEY NOT NULL,
+  ela_adaptive_skill_practice_content_id uuid REFERENCES ela_adaptive_skill_practice_content NOT NULL,
+  type text NOT NULL,
+  content text NOT NULL,
+  caption text,
+  position integer NOT NULL,
+  UNIQUE (ela_adaptive_skill_practice_content_id, position),
+  CONSTRAINT valid_ela_adaptive_skill_practice_paragraph_position CHECK (position >= 0)
+);
+```
+
+Do this:
+
+```diff
+CREATE TABLE ela_adaptive_skill_practice_paragraphs (
+  id integer PRIMARY KEY NOT NULL,
+- ela_adaptive_skill_practice_content_id uuid REFERENCES ela_adaptive_skill_practice_content NOT NULL,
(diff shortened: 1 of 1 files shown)


pbrisbin (approved):
:clap: well-said.

eborden (approved):

jdreaver (dismissed):
I'm a bit confused by the foreign key example. If the column name is `assignment_id`, then doesn't `mathStandardAssignmentId` not even work? Don't we have to call it `assignmentId`? (I totally think I'm wrong here since this did in fact work haha, I'm mostly looking for clarification.)

Maybe the convention should be: use column names in Haskell that exactly match the column name in SQL. Talking about _removing_ a prefix kinda implies we started with one somehow (which makes sense for legacy entities), but if we just say "use the same column name, just in camelCase", we don't even have to talk about prefixes.

cdparks:
@jdreaver yeah, this is a little confusing because it's really both a Haskell and DB issue. We have some tables where the DB column _is_ prefixed like this, and we've also prefixed the Haskell entity field name, e.g.:

```sql
CREATE TABLE public.ela_adaptive_skill_practice_paragraphs (
    id integer NOT NULL,
    ela_adaptive_skill_practice_content_id uuid NOT NULL,
    type text NOT NULL,
    content text NOT NULL,
    caption text,
    "position" integer NOT NULL,
    CONSTRAINT valid_ela_adaptive_skill_practice_paragraph_position CHECK (("position" >= 0))
);
```

and

```haskell
share [mkPersist sqlSettings, mkMigrate "migration"] [persistLowerCase|
ElaAdaptiveSkillPracticeParagraph sql=ela_adaptive_skill_practice_paragraphs
  elaAdaptiveSkillPracticeContentId ElaAdaptiveSkillPracticeContentId
  type ElaParagraphType
  content Text
  caption Text Maybe
  position Natural
  UniqueElaAdaptiveSkillPracticeParagraphElaAdaptiveSkillPracticeContentIdPosition  elaAdaptiveSkillPracticeContentId position
  deriving Show Eq Generic
|]
```

This is bad because `AdaptiveSkillPractice` isn't right. It should be `SkillsPractice`. However, because of this over-prefixing, I can't quick-fix it because the extra prefixing is now part of the JSON API.

Maybe this is a better motivating example?

cdparks:
Also, there's definitely a switch in the quasiquoter syntax to use a different field name than what the DB column actually looks like.

jdreaver:
Ah that example makes more sense. From how this rule is written it looks like a primarily Haskell concern though instead of a schema naming concern. I would phrase any rules here like this:
- _In our schema_, don't over-prefix names (i.e. `ela_adaptive_skill_practice_content_id` should be `content_id` or whatever it should be)
- _In our Haskell entities_, use the exact same column names as our schema, just in camelCase instead of snake_case

This gives us the naming we want _and_ keeps out Haskell names consistent with our schema, so there are no surprises between them. To be clear, I'm totally in agreement that this over-prefixing is bad. I'm just trying to simplify the rules we use in the guide, and I also think saying "just use the SQL column names in camelCase" is generally what we want anyway.

Having our DB column names be part of the JSON schema is unfortunate :cry:

jdreaver (approved):
Thanks @cdparks! LGTM

PR #25, opened 2018-07-27
Avoid MonadFail best practice

pbrisbin:
I got bit by this in a recent task, luckily caught by tests. I'm pretty sure
this is a best-practice some of us follow, so I'm proposing we document it.

I tried to provide accurate details in the justification, verified in ghci, but
I don't feel like an expert in this area so please let me know if something's
inaccurate or even if you think I should just remove those details as
superfluous.

I know @eborden has some opinions here, and I'm 100% happy to defer if he wants
to rewrite this himself.

### DIFF
--- a/haskell-best-practices.md
+++ b/haskell-best-practices.md
@@ -13,6 +13,67 @@
      x `divEx` d = x / d
  ```
 
+### Avoid `MonadFail`
+
+Don't write our own functions using `MonadFail`/`fail`.
+
+**Bad**:
+
+```hs
+data Email = Email
+
+parseEmail :: MonadFail m => Text -> m Email
+parseEmail = undefined
+
+instance FromJSON Email where
+  parseJSON = withText "Email" parseEmail
+```
+
+This looks convenient:
+
+```hs
+parseEmail "..." :: Maybe Email -- fail as Nothing
+parseEmail "..." :: IO Email -- fail as throw
+```
+
+It would be reasonable to think you can also:
+
+```hs
+parseEmail "..." :: Either String Email -- fail as Left
+```
+
+And that *may* work, but it depends on a lot of implicit context and has
+surprising *run-time* failure modes:
+
+For the above to work, there would need to be a `MonadFail` instance for `Either
+String` in scope that defines `fail = Left`. That itself is not hugely
+problematic, but *when there isn't* the resulting behavior further depends on if
+you've `import`ed `Control.Monad.Fail` or not. If you *have* you'll get a
+compile-time "no instance for (MonadFail (Either String))" error. Nice. But if
+you *haven't*, you'll implicitly get `Monad(fail)`, which is a run-time `error`
+call for `(Either e)`. Not good.
+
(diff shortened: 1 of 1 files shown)


jdreaver (approved):
:heart: concrete monad stacks unless there is a clear abstraction needed that warrants a type class constrained monad.

haskell-best-practices.md
```diff
@@ -13,6 +13,67 @@
      x `divEx` d = x / d
  ```
 
+### Avoid `MonadFail`
+
+Don't write our own functions using `MonadFail`/`fail`.
```
5outh: Should `fail` be considered a restricted keyword in our hlint configuration?

haskell-best-practices.md
```diff
@@ -13,6 +13,67 @@
...
+
+**Bad**:
```
5outh: This is super nitpicky, but I think this might read better if the sections were sequenced as follows:

1. Bad example
2. Good example
3. Justification

5outh (approved):
In general in favor of this; I agree with David that we should be preferring concreteness over abstraction in this case.

haskell-best-practices.md
eborden: Sadly there are some libraries which utilize `MonadFail` as their only point of failure. Ex: `cassava`.

eborden: Can we change this to:
> Don't write functions constrained by `MonadFail`.

haskell-best-practices.md
```diff
@@ -13,6 +13,67 @@
...
+```hs
+parseEmail "..." :: Maybe Email -- fail as Nothing
+parseEmail "..." :: IO Email -- fail as throw
+```
+
+It would be reasonable to think you can also:
+
+```hs
+parseEmail "..." :: Either String Email -- fail as Left
+```
```
eborden: How about something like:

`MonadFail` over `Either` is a common foot-gun. We might expect `fail` to return `Either String a`, but instead it throws an `error`. There are many other surprising and problematic instances of `MonadFail` that make this abstraction dangerous. Writing functions that abstract over `MonadFail` opens us up to these surprising and hard to reason about errors.

haskell-best-practices.md
```diff
@@ -13,6 +13,67 @@
...
+
+Don't write our own functions using `MonadFail`/`fail`.
+
+**Bad**:
+
+```hs
+data Email = Email
+
+parseEmail :: MonadFail m => Text -> m Email
+parseEmail = undefined
```
eborden: can we use `_` instead of `undefined`. Or even write a simple body?

```
parseEmail :: MonadFail m => Text -> m Email
parseEmail
  | isEmail txt = Email txt
  | otherwise = fail "that isn't an email!"
```

eborden (commented):
I'm in favor of this, just a few thoughts.

pbrisbin:
Updated

eborden (approved):

cdparks (approved):

PR #27, opened 2018-09-18
Specify naming guide for request/response types.

pbrisbin:
NOTE: This was an attempt to codify what we do most-commonly at this
time, not an attempt at finding an objectively good guide. If folks want
to propose an objectively good guide, have at it.

We are currently in violation of this guide because we're inconsistent
in casing the `Method` suffix, but these types are very easy to identify
and rename, so I plan to resolve things as soon as this is accepted.

### DIFF
--- a/naming-conventions.md
+++ b/naming-conventions.md
@@ -229,8 +229,8 @@ This will produce the following JSON:
 ### Request/Response types
 
 For complicated `Handler` requests or responses, where we use a custom type
-instead of domain object, such types should be named `{Resource}{Method}` where
-`Resource` matches the route and `Method` is Sentence-case.
+instead of a domain object, such types should be named `{Resource}{Method}`
+where `Resource` matches the route and `Method` is Sentence-case.
 
 ```hs
 -- Good
--- a/naming-conventions.md
+++ b/naming-conventions.md
@@ -226,6 +226,53 @@ This will produce the following JSON:
 }
 ```
 
+### Request/Response types
+
+For complicated `Handler` requests or responses, where we use a custom type
+instead of an `Entity` directly, such types should be named `{Resource}{Method}`
+where `Resource` matches the route and `Method` is Sentence-case.
+
+```hs
+-- Good
+data SchoolGet
+  -- ...
+
+instance ToJSON SchoolGet where
+  -- ...
+
+getSchoolR :: SchoolId -> Handler Value
+getSchoolR schoolId = do
+  school <- get404 schoolId
+  sendStatusJSON status200 SchoolGet {- ... -}
+
+data SchoolsPost
+  -- ...
+
+instance FromJSON SchoolsPost where
+  -- ...
+
+postSchoolsR :: Handler Value
+postSchoolsR = do
+  SchoolsPost {..} <- requireJsonBody
+  -- ...
+
+-- Bad
+data GetSchool
+  -- Out of order (though gramatically attractive)
+
+data SchoolGET
+  -- Wrong case
+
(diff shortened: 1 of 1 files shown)


naming-conventions.md
```diff
@@ -226,6 +226,53 @@ This will produce the following JSON:
...
+data SchoolsPost
+  -- ...
+
+instance FromJSON SchoolsPost where
+  -- ...
+
+postSchoolsR :: Handler Value
+postSchoolsR = do
+  SchoolsPost {..} <- requireJsonBody
+  -- ...
```
cdparks, line 257: Odd (for us) case - if you have a response for a `POST`, do you do `SchoolsPostRequest` and `SchoolsPostResponse`?

naming-conventions.md
```diff
@@ -226,6 +226,53 @@ This will produce the following JSON:
 }
 ```
 
+### Request/Response types
+
+For complicated `Handler` requests or responses, where we use a custom type
+instead of an `Entity` directly, such types should be named `{Resource}{Method}`
```
cdparks: I also think another acceptable exception is data types that have some obvious name unrelated to being a `GET` response, e.g `ElaReadingReport`. Especially if you're doing a bunch of pure logic to generate one, the fact that it's being used as a `GET` response doesn't feel important enough to reflect in the naming to me.

cdparks (approved):

naming-conventions.md
pbrisbin: Yeah, I thought of that and purposely didn't talk about. Partly because we have no cases (that I know of) and partly because I'm not sure what to prescribe. If you'd like to see a sentence in here about that (either prescribing something or just saying that it's left unspecified) please let me know.

naming-conventions.md
pbrisbin: That's a good point. I agree we should soften the language to not let this guide preclude making a proper domain object that can also serve as the request/response type.

Maybe instead of saying:

> a custom type instead of an `Entity` directly, ...

We can say

> a custom type instead of an existing domain object, ...

In the `Report` case, I would call that a generally-useful domain object that just happens to only be used to support a response at the moment.

naming-conventions.md
cdparks: Eh, probably better to leave it unspecified. I suspect if we do ever have `POST`s that return something, the result will be an `Entity` potentially `WithMetadata`

naming-conventions.md
cdparks: Yeah, I like the domain object language

naming-conventions.md
```diff
@@ -226,6 +226,53 @@ This will produce the following JSON:
 }
 ```
 
+### Request/Response types
+
+For complicated `Handler` requests or responses, where we use a custom type
+instead of domain object, such types should be named `{Resource}{Method}` where
+`Resource` matches the route and `Method` is Sentence-case.
+
+```hs
```
cdparks: Aw nuts I've been typing out `haskell` for code blocks this whole time

naming-conventions.md
```diff
@@ -226,6 +226,53 @@ This will produce the following JSON:
 }
 ```
 
+### Request/Response types
+
+For complicated `Handler` requests or responses, where we use a custom type
+instead of domain object, such types should be named `{Resource}{Method}` where
```
cdparks: Nit `a domain object` or `domain objects`

cdparks (approved):

eborden (approved):
No qualms here.

PR #30, opened 2018-10-31
Add Haskell OSS document

pbrisbin:
Some of this may be a bit contentious, but I figured I'd throw it out there as
something concrete to discuss.


haskell-open-source.md
```diff
@@ -0,0 +1,35 @@
+## Best Practices
+
+1. If authoring a distinct package of reasonable size and generic utility,
+   prefer starting a separate, public repository utilized as a Git dependency
+
+When working in a public repository:
+
+1. Keep a linear Git history with rich, meaningful commit messages
+
+   Configure GitHub to only allow Squash and Rebase strategies.
```
eborden, line 10: You really are gunning for controversy :smile: 

:+1:

haskell-open-source.md
```diff
@@ -0,0 +1,35 @@
+## Best Practices
+
+1. If authoring a distinct package of reasonable size and generic utility,
+   prefer starting a separate, public repository utilized as a Git dependency
```
eborden, line 4: The line here seems fuzzy. I feel like things often start out as small utilities and grow into something that should be made into a library/module. Where that happens I have no idea.

eborden (approved):

haskell-open-source.md
pbrisbin: I'm happy to change it (or remove it). I'm trying to put just a little bit of pressure to move away from the failure mode of starting with a sub-directory and never extracting it. Maybe we don't care about that risk?

akurilin (approved):

haskell-open-source.md
eborden: :man_shrugging:

5outh (approved):

haskell-open-source.md
jdreaver: heh I love merge commits. It's much easier to see a merge commit like "Merged pull request XXX" so I know where a chunk of code came from vs an uninformative cleanup commit like "fixed linting issues" that inevitably come at the end of a PR :man_shrugging: Feel free to keep this but I don't think its necessary at all for an OSS guide.

jdreaver (approved):

haskell-open-source.md
pbrisbin: > vs an uninformative cleanup commit like "fixed linting issues" that inevitably come at the end of a PR

Yeah, that's been an unfortunate risk, but it's why Squash is there: if one doesn't want to take the time to rebase out any fixup commits, they can just choose Squash.

It's no harder than hitting Merge, it creates one commit, and it'll at least use the PR title and list of commit details instead of an uninformative message like "Merged pull request XXX", which _really_ frustrated me that other day when I was trying to find out what merges could've impacted our logging setup...

I think that's still everything you love about merge commits. What actually makes merge commits unique (preserving an _honest_ history instead of an _ideal_ history) isn't what I think you're arguing for.

jdreaver: Yeah I think I focused on rebase too much here. I think squash would be better than both merge commits and rebase now that you pointed it out. :+1: (I'm not taking a hard position here. If the team prefers any one of these I'm fine with it.)

eborden: Yeah, sounds like you are more arguing against "bad" commits than against rebase.

jdreaver: Not really. I don't think satisfying a linter in its own commit is necessarily bad. If I had a PR with 5 commits and each one introduced a tiny bit of linting failure, I wouldn't want to go and edit each of those 5 commits. I would probably just introduce a new one at the end. 

When I see a commit summary in CircleCI, Jenkins, or even the git log, I like that merge commits tell me exactly which PR to go to for more context on the entire change. Squashes being self contained helps too. With a rebase merge that becomes harder and requires more digging.

pbrisbin: > With a rebase merge that becomes harder and requires more digging.

You probably know this, and still consider it "more digging" but if you open the commit on GitHub (clicking through from Circle or Jenkins), GH has a link to the PR in the header.

eborden: My problem with squash is that it ends up diluting a clean commit stream. I attempt to make each of my commits atomic and include a description of what/why everything was done. When you squash it dilutes the change set. Now a specific commit message will be interspersed with many others and it will be hard to tell which what/why is applicable.

pbrisbin: Let's be careful here. This conversation tends to drift towards either-or, but we can all be happy very easily:

- If you like to rebase to a detailed multi-commit PR with good messages, please do so
- If you don't, please just use Squash instead of Merge

eborden: Oh totally agree, I'm just shaving yaks at this point.

pbrisbin:
Merging on majority. Chris is out for a few days.

PR #31, opened 2018-11-07
Add guidelines for MonoTraversable and OverloadedLists

akurilin:


haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
 f (ThingB b) = if b then 1 else 2
 ```
 
+## Avoid `OverloadedLists` and `MonoTraversable`
+
+`OverloadedLists` work from a typelcass `IsList`:
```
cdparks: Spelling, probably "`OverloadedLists` depend on a typeclass `IsList`"

haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
...
+
+```
+class IsList l where
+type Item l
+fromList  :: [Item l] -> l
+toList    :: l -> [Item l]
```
cdparks: Probably wanna indent everything under `class IsList l where`

haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
...
+```
+  
+The `mono-traversable` package defines a series typeclasses:
```
cdparks: "a series of typeclasses"

haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
...
+
+```
+type family Element mono
+
+class MonoFoldable mono where
+ofoldMap :: Monoid m => (Element mono -> m) -> mono -> m
```
cdparks: Mo' indentin', mo' bettah

haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
...
+
+The problem gets deeper when we realize that `mono-traversable` is a nearly useless abstraction for our use-case. There are exactly *ZERO* uses of `omap` in our codebase. We love `map`, `foldMap`, `traverse`, `for`, etc. These are infinitely useful, but we are never for example traversing over the characters of a `Text` type, which is `mono-traversable`'s primary inspiration. If our use case demanded a lot of string munging, then maybe it would be useful. That just isn't the case.
+
+The fact is that `mono-traversable` is not an abstraction that targets our use-case, it decreases type inference, increases cryptic type errors, and increases boilerplate.
+
+After all that you might say, but `FrontRow.Model.Assignment` uses type families and associated types. And you'd be right. That use of type families and associated types was very purposefully designed to abstract over a common use case. It has known issues of type inference and hairy type errors, but it is a matter of ROI. `FrontRow.Model.Assignment` returns on its complexity investment, `mono-traversable` does not.
```
cdparks: Assignment families also use injective type families, which have better type inference than non-injective type families (the default), but that may not be worth working into this.

haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
 f (ThingB b) = if b then 1 else 2
 ```
 
+## Avoid `OverloadedLists` and `MonoTraversable`
+
+`OverloadedLists` work from a typelcass `IsList`:
+
+```
```
pbrisbin: Can we get some syntax highlighting on these?

```
```hs
```haskell
```

Both will work

haskell-best-practices.md
pbrisbin: I'd probably avoid Freckle-code-base-specific examples here. Dropping this paragraph doesn't diminish the impact here IMO.

haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
...
+type family Element mono
+
+class MonoFoldable mono where
+ofoldMap :: Monoid m => (Element mono -> m) -> mono -> m 
+...
+```
+  
+You can see that both of these typeclasses utilize type families. `IsList` has an associated type `Item` and `MonoFoldable` uses an open type family `Element`. They are nearly identical in their utilization of these techniques. I'll focus mainly on `mono-traversable`, but arguments against it are very transferable to `OverloadedLists`.
```
pbrisbin: I'd love to see these hard-wrapped as we do *mostly* consistently in our markdown (this document isn't a great example). But I get that it's a pain without a properly-configured editor, so consider this non-blocking.

pbrisbin (approved):

haskell-best-practices.md
cdparks: I know this is from a slack comment, which is probably why it's somewhat colloquial. How about

> Each of these typeclasses rely on type families:
> * `IsList` has an associated type `Item` for representing the element of the list
> * `MonFoldable` uses an open type family `Elem` for representing the element of `mono`

You can probably lead into the next paragraph with

> Unfortunately, type families do not have great type inference. Furthermore, `mono-traversable` exports a number of replacements for common `Prelude` functions which rely on these type families.

haskell-best-practices.md
```diff
@@ -396,6 +396,37 @@ f (ThingA a) = a == 0
...
+
+The problem is that type families do not have great type inference. The problem is compounded by the fact that `mono-traversable` supplants a large part of the prelude with functions that have bad type inference. We end up getting tons of cryptic type errors in places where we'd expect type inference to take hold. That leads to confusion and littering type annotations around the codebase. So we don't like them.
```
cdparks: > So we don't like them.

I don't know if this is necessary, especially since it's not obvious if we're talking about `mono-traversable` or type families in general :smile:

cdparks (changes requested):
I dunno how much review you want. Are you wanting to copy/paste this in here, or are you looking for copy-editing?

PR #32, opened 2019-03-04
Add a very basic API style guide

5outh:
As the title says; adds a couple of basic rules about the style our API should uphold.

### DIFF
--- a/api-style.md
+++ b/api-style.md
@@ -1,0 +1,39 @@
+# API Style Guide
+
+## Casing
+
+- Object fields should be `camelCase`
+- Enum identifiers should be `snake_case`
+
+bad
+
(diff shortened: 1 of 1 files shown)


pbrisbin (changes requested):
I think we could find a way to fold this into the more general naming document. We've already started a little bit here: https://github.com/frontrowed/guides/blob/master/naming-conventions.md#generating-enums-from-haskell-for-javascript

Sorry if this makes this more annoying. Even if we want to sprout a new file, I'm not sure "API Style" is right, since these JSON rules should really apply to all Haskell JSON instances (IMO), such as Job payloads.

eborden:
Closing this since it hasn't been touched in a long time.

PR #37, opened 2019-04-24
The deriving clause should be on a new line

adomokos:
@pbrisbin suggested updating our style guide with modification. What do you think?

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -270,7 +270,8 @@ data SomeRecord'
   = SomeRecord'
   { someField :: Int
   , someOtherField :: Double
-  } deriving (Eq, Show)
+  }
+  deriving (Eq, Show)
 
 -- Bad
 data SomeSum = FirstConstructor


pbrisbin (approved):
This is what I thought our existing style was, and is my own preference. The reasoning is for how it interacts with multiple `deriving` clauses, which we also like and use a lot.

With same-line `deriving`, adding another clause is either a rough diff:

```diff
- } deriving stock Eq
+ }
+ deriving stock Eq
+ deriving newtype ToJSON
```

Or left looking odd

```hs
} deriving stock Eq
deriving newtype ToJSON
```

With next-line, both problems are solved

```diff
 }
 deriving stock Eq
+deriving newtype ToJSON
```

It's basically the same argument as for any "start this thing on the next line" guide we have (types, tuples, lists, arguments, applicatives).

cdparks (approved):

eborden (approved):
Agreed on the preference.

PR #39, opened 2019-07-11
Draft API conventions

pbrisbin:
Pulling together various conversations I've had about conventions we want to have in place as we build out the V3 API.

### DIFF
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -43,7 +43,7 @@ For example:
 ]
 ```
 
-Here, a School filter would be `school.id`.
+Here, a School filter would be `school.id`: `/3/teachers?school.id=1,2,3`.
 
 ```
 /3/students
@@ -58,7 +58,7 @@ Here, a School filter would be `school.id`.
 ]
 ```
 
-But here, it would be `schools.id`.
+But here, it would be `schools.id`: `/3/students?schools.id=1,2,3`.
 
 If a filter exists that is not for an attribute present in the response, the
 name can be inferred by what it would look like if it were.
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -95,7 +95,7 @@ than our multi-id path-pieces of today.
 - Include related resources in abbreviated form, Frontend almost always wants
   names:
   - `students: [{id:, firstName:, lastName:}]`
-  - `school: [{id:, name:}]`
+  - `school: {id:, name:}`
 
 ## Status Codes
 
@@ -120,16 +120,17 @@ than our multi-id path-pieces of today.
   - There's no need to `V3`-prefix everything
   - Things can "start over" when there's the "single below a list" situation,
     e.g. `TeacherP x $ TeacherR` instead of `TeachersP $ TeachersTeacherP x $
-    TeachersTeacherR`, which is a bit buffalo-buffalo.
+    TeachersTeacherR`, which is a bit buffalo-buffalo
+  - The "start over" caveat does not apply to path naming (i.e. the above
+    example will still have the module `Teachers/Teacher.hs`)
 
 - Name modules to directly match parents and routes, which implies one Handler
   module per route (see example)
 
   **Caveats**:
 
-  - In reality, we need to prevent collisions between versions through additional
-    prefixing of route constructures. We ignore this in the guide and do it only
-    where needed (e.g. in the constructor, but not the module name).
+  - In reality, we need to prevent collisions between versions through
+    additional prefixing of route constructors. We ignore this in this guide.
 
 Example:
 
@@ -146,6 +147,29 @@ Example:
 ## Module Structure
 
 - Prefer in-module request/response types
+
+  - Request type naming: `{Resource}{Method}Body`
+  - Response type naming: `{Resource}{Method}`
+  - NOTE: list responses are `[{Resource}{Method}]`, not `{Resources}{Method}`
+
+  ```hs
+  -- Good
+  data TeacherGet
+
+  getTeachersR :: Handler [TeacherGet]
+  getTeachersR = undefined
+
+  data TeacherPostBody
+
+  -- Bad
+  data Teachers
+
+  getTeachersR :: Handler Teachers
+  getTeachersR = undefined
+
+  data PostTeacher
+  ```
+
 - Prefer in-module data-access functions, until they require sharing
 - Order your module as:
   - Request/response type(s)
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -9,8 +9,8 @@ choose what data to return.
   - `/3/students?schools.id=x,y,z`
   - `ApiClient.fetch("/3/students", { "schools.id": me.schoolIds })`
 - Bad:
-  - `/3/school-admins/me/schools`
-  - `ApiClient.fetch("/3/school-admins/me/schools")`
+  - `/3/school-admins/me/students`
+  - `ApiClient.fetch("/3/school-admins/me/students")`
 
 ## Parameters
 
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -100,19 +100,13 @@ than our multi-id path-pieces of today.
 
 ---
 
-## Request Shape
-
-- Request bodies can vary by use, but this should be avoided if possible
-
-*TODO*: more to say here?
-
 ## Response fields
 
 - Always include `id`s
 - Within reason, choose attributes directly for the work that is motivating the
   endpoint
-- Include related resources in abbreviated form, Frontend almost always wants
-  names:
+- Include related resources in abbreviated form when Frontend needs extra data
+  about them. For example:
   - `students: [{id:, firstName:, lastName:}]`
   - `school: {id:, name:}`
 
@@ -212,16 +206,22 @@ See `/3/teachers` as a good example.
 
 Do not export your request/response types for use in tests. Instead, re-build
 JSON `Value`s with `object` This ensures you don't have a (de)serialization bug
-that passes the tests because it's used in both places. It also makes it clear
-how to move to `aeson-lens`-based assertions (which are more robust) when that
-makes sense.
+that passes the tests because it's used in both places.
 
 ```hs
 -- Good
 body <- getJsonBody
-body `shouldBe` [object ["id" .= teacherId]]
+body `shouldMatchList` [aesonQQ|
+  [ { id: #{teacherId}
+    }
+  ]
+|]
 
--- Doesn't fail on unrelated attribute changes
+-- Or
+body <- getJsonBody
+body `shouldMatchList` [object ["id" .= teacherId]]
+
+-- This doesn't fail on unrelated attribute changes
 body <- getJsonBody @Value
 body
   ^.. _Array
@@ -232,5 +232,5 @@ body
 
 -- Bad
 body <- getJsonBody
-body `shouldBe` [TeacherGet teacherId]
+body `shouldMatchList` [TeacherGet teacherId]
 ```
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -90,9 +90,9 @@ This is because:
 ## Status Codes
 
 - 201 for creation
+- 202 for async jobs (with `Link` header)
 - 204 for no content (i.e. our `Empty` responses)
 - 400s MUST follow our `ValidationError` machinery
-- 202 for async jobs (with `Link` header)
 
 ## Pagination
 
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -115,6 +115,7 @@ than our multi-id path-pieces of today.
 - 201 for creation
 - 204 for no content (i.e. our `Empty` responses)
 - 400s MUST follow our `ValidationError` machinery
+- 202 for async jobs (with `Link` header)
 
 ## Pagination
 
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -159,7 +159,12 @@ Example:
   getTeachersR :: Handler [TeacherGet]
   getTeachersR = undefined
 
+  data TeacherPost
   data TeacherPostBody
+
+  postTeacherR :: Handler TeacherPost
+  postTeacherR = do
+    body <- requireJsonBody @TeacherPostBody
 
   -- Bad
   data Teachers
@@ -168,6 +173,10 @@ Example:
   getTeachersR = undefined
 
   data PostTeacher
+
+  postTeacherR :: Handler (Entity Teacher)
+  postTeacherR = do
+    body <- requireJsonBody @PostTeacher
   ```
 
 - Prefer in-module data-access functions, until they require sharing
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -193,6 +193,15 @@ makes sense.
 body <- getJsonBody
 body `shouldBe` [object ["id" .= teacherId]]
 
+-- Doesn't fail on unrelated attribute changes
+body <- getJsonBody @Value
+body
+  ^.. _Array
+  . traverse
+  . key "id"
+  . _JSON
+  `shouldMatchList` [teacherId]
+
 -- Bad
 body <- getJsonBody
 body `shouldBe` [TeacherGet teacherId]
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -1,0 +1,175 @@
+# Haskell API Conventions
+
+## Stateless
+
+Prefer stateless endpoints. Sessions are used to authorize data access, but not
+choose what data to return.
+
+- Good:
+  - `/3/students?schools.id=x,y,z`
+  - `ApiClient.fetch("/3/students", { "schools.id": me.schoolIds })`
+- Bad:
+  - `/3/school-admins/me/schools`
+  - `ApiClient.fetch("/3/school-admins/me/schools")`
+
+## Parameters
+
+We want to :heart: query parameters as a way to keep the number of distinct
+endpoints lower by supporting more complex filtering combinations within each
+endpoints.
+
+- Good: `/3/students?schools.id=x,y,z&teachers.id=a,b,c`
+- Bad:
+  - `/3/schools/:id/students`
+  - `/3/teachers/:id/students`
+  - `/3/schools/:id/teachers/:id/students`
+  - etc, forever
+
+### Naming
+
+- If filtering on an attribute of the main resource: `{attribute}={value}`
+- If filtering on an attribute of a related resource:
+  `{resource}.{attribute}={value}`
+- If one thing has many of another, pluralize the many
+
+Examples, for a `/teachers` endpoint:
+
+- `?name=`, because a Teacher has one name
+- `?school.id=`, because School is a relation, and a Teacher has only one
+- `?students.id=`, because Student is a relation, and a Teacher has many
+- `?grades=`, direct and has many
+
+Note that pluralization can occur in either position. For example, Teachers have
+many Students, and (if we assume for the sake of this of this example) Students
+have many levels, then you would filter Teachers by their Students' levels with
+`students.levels=`.
+
+### Semantics
+
+Prefer `IN` equality semantics.
+
+- Good: `{attribute}={value1},{value2},{value3},...`
+- Bad: `{attribute}={value}`
+
+This is because:
+
+- It's just as easy to support as direct equality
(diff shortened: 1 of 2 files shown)


haskell-api.md
```diff
@@ -0,0 +1,217 @@
...
+  - `/3/schools/:id/students`
+  - `/3/teachers/:id/students`
+  - `/3/schools/:id/teachers/:id/students`
+  - etc, forever
+
+### Naming
+
+- If filtering on an attribute of the main resource: `{attribute}={value}`
+- If filtering on an attribute of a related resource:
+  `{resource}.{attribute}={value}`
```
eborden: It seems like we consider foreign keys to be considered an attribute on a related resource. i.e.

```sql
CREATE TABLE teacher
  ( schoolId int REFERENCES schools)
```

Bad
```
GET /3/teachers?schoolId=123
```

Good
```
GET /3/teachers?schools.id=123
```

haskell-api.md
```diff
@@ -0,0 +1,217 @@
...
+- If one thing has many of another, pluralize the many
+
+Examples, for a `/teachers` endpoint:
+
+- `?name=`, because a Teacher has one name
+- `?school.id=`, because School is a relation, and a Teacher has only one
```
eborden: I think it is easier to consider relations always plural. A teachers endpoint may pull back many teachers and they may have many `schools`. Relations are always plural is much simpler than anything else.

eborden (commented):
Took a quick pass over some of it, more to come later.

haskell-api.md
pbrisbin: I did some more thinking on this because I agree it's got a downside in complexity as written.

However, I think it's actually simpler to just say it should match a path in the response:

```
/teachers
[ { id:
  , school: { id: }
  }
]
```

Therefore, `/teachers?school.id=`.

```
/students
[ { id:
  , schools: [{ id:, ... }]
  }
]
```

Therefore, `/students?schools.id=`.

This is the same exact outcome, but a lot simpler to state as a guide. The only gap is when you are filtering by something not in the response, which we will likely support, but can (IMO) be inferred by how it *would* look in a response.

WDYT?

haskell-api.md
```diff
@@ -0,0 +1,217 @@
...
+  - Repeat if more than one handler
+  - Internal functions
+- Document your endpoints
+
+See `/3/teachers` as a good example.
+
+## Testing
+
+Do not export your request/response types for use in tests. Instead, re-build
+JSON `Value`s with `object` This ensures you don't have a (de)serialization bug
```
jdreaver, line 186: :clap:

haskell-api.md
```diff
@@ -0,0 +1,217 @@
...
+
+  data PostTeacher
+
+  postTeacherR :: Handler (Entity Teacher)
+  postTeacherR = do
+    body <- requireJsonBody @PostTeacher
+  ```
+
+- Prefer in-module data-access functions, until they require sharing
+- Order your module as:
```
jdreaver, line 174: :clap:

haskell-api.md
eborden: Agreed, that is a clearer and more intuitive criteria since it matches the resource and the filter.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+  `schools.id` in the single endpoint and the Frontend can just switch
+
+Therefore, I find the length concerns for `IN`-queries to be less problematic
+than our multi-id path-pieces of today.
+
+---
+
+## Request Shape
+
+- Request bodies can vary by use, but this should be avoided if possible
```
cdparks: I'm not following. Does this mean request bodies for the same route can vary based on the _user_?

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+*TODO*: more to say here?
+
+## Response fields
+
+- Always include `id`s
+- Within reason, choose attributes directly for the work that is motivating the
+  endpoint
+- Include related resources in abbreviated form, Frontend almost always wants
+  names:
```
cdparks: > Frontend almost always wants names

This seems like a recent phenomenon. Can we get a frontender to comment? @vmerino @joris974  For some resources, we're likely to have them laying around, so doing a simple frontend join isn't too bad.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+  This means we won't break Frontend when the tree expands over time
+
+- Fully name-space all parents and routes (see example)
+
+  **Caveats**:
+
+  - There's no need to `V3`-prefix everything
+  - Things can "start over" when there's the "single below a list" situation,
+    e.g. `TeacherP x $ TeacherR` instead of `TeachersP $ TeachersTeacherP x $
```
cdparks, line 113: Probably `TeachersP x $ TeacherR`

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+See `/3/teachers` as a good example.
+
+## Testing
+
+Do not export your request/response types for use in tests. Instead, re-build
+JSON `Value`s with `object` This ensures you don't have a (de)serialization bug
+that passes the tests because it's used in both places. It also makes it clear
+how to move to `aeson-lens`-based assertions (which are more robust) when that
+makes sense.
```
cdparks: After fiddling with `aeson-lens` this morning, I'm less enthusiastic about this. I find `aesonQQ` plus a reasonably loose predicate a lot more readable.

cdparks (approved):

haskell-api.md
pbrisbin: This was just meant to capture some existing case such as the session-creation route that have different shape by product. I can just remove it, it's not important.

haskell-api.md
pbrisbin: It is recent in the Roster Management work, I can just nix this part mentioning `name`. It's just an example of why an abbreviated resource can-be/has-been useful.

haskell-api.md
pbrisbin: Not if you're following the "expand every parent to start" guide.

```
-- Good
/teachers TeachersP:
  / TeachersR
  /#TeacherId TeacherP:
    / TeacherR GET

-- Bad
/teachers TeachersP:
  / TeachersR
  /#TeacherId TeacherR GET
```

Because *Good* won't break frontend when things expand:

```diff
 /teachers TeachersP:
   / TeachersR
   /#TeacherId TeacherP:
     / TeacherR GET
+    /whatever TeacherWhateverP:
+     / TeacherWhateverR GET
```

But *Bad* does

```diff
 /teachers TeachersP:
    / TeachersR
-   /#TeacherId TeacherR GET
+   /#TeacherId TeacherP:
+     / TeacherR GET
+     /whatever TeacherWhateverR GET
```

Unless you are OK with things going this way, which devolves into rampant improvisation of either style and/or can't support some kind of later changes either:

```diff
 /teachers TeachersP:
    / TeachersR
    /#TeacherId TeacherR GET
+   /#TeacherId/whatever TeacherWhateverR GET
```

haskell-api.md
pbrisbin: FWIW, I had the same experience, and I know it's not convincing, but after using it in anger over three features I'm absolutely in love. It's an annoying journey, but on the other side has been bliss for me.

I like it because it's so flexible that I can make all sorts of specific assertions that clearly show my intent in the test -- in a consistent way for any assertion. I don't have to do things differently if I want whole-body vs targeted, it's all the same thing to match on a sub-doc, order of keys, attribute, etc.  To me that's more readable on a macro level, not just in the `aesonQQ`-vs-`object` comparison.

That said, I don't feel strongly about pushing against `aesonQQ` -- it's a subjective "me" thing about how I prefer to write my tests. This guide doesn't preclude `aesonQQ` at all, so I can just remove this point too.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+But here, it would be `schools.id`.
+
+If a filter exists that is not for an attribute present in the response, the
+name can be inferred by what it would look like if it were.
+
+### Semantics
+
+Prefer `IN` equality semantics.
```
5outh, line 68: Just to be clear, we're not talking about _not supporting_ operating on a single entity by attribute/value, but _always supporting_ operating on multiples?

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+**Aside about URL length limits**: We don't know at what size this will be a
+problem. Standards-wise, there is no limit, but browsers and servers do have
+limits and it's recommended to use 2000 as a reference point [ref].
+
+[ref]: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
+
+However, I will say this:
+
+- Our current `/foo/:id,:id,:id/bar` does bump into issues of length and it is
```
5outh: More filtering parameters could alleviate this problem.

Like, if we are fetching multiple `course`s by id, a GET request for them might look like this:

```
GET /2/courses/1,2,3,4,5,...
```

but typically there is structure around those courses. For example, this might be fetching all Courses in a (set of) School(s), and could be rewritten using a filter parameter on the route:

```
GET /3/courses?schools.id=1,2
```

which shortens the URL considerably.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+  annoying to resolve because it means a new *route* to find a natural way to
+  group the massive list (i.e. all students in a school will mean changing
+  `/students/...` to `schools/:id`)
+- Doing the same thing in a `GET` param will be much nicer to handle when we
```
5outh: (maybe this :point_up: is what you were saying here)

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+than our multi-id path-pieces of today.
+
+---
+
+## Response fields
+
+- Always include `id`s
+- Within reason, choose attributes directly for the work that is motivating the
+  endpoint
+- Include related resources in abbreviated form when Frontend needs extra data
```
5outh, line 85: So `students` (e.g.) can mean multiple things depending on the endpoint? I'm ok with this as long as it's written down.

haskell-api.md
5outh: <sup>it would be nice if route generation worked on routes, not their haskell representation :grimacing:</sup>

Another option is to flatten completely, which i'm quietly a fan of:

```haskell
/teachers TeachersR GET 
/teachers/#TeacherId TeacherR GET
/teachers/#TeacherId/whatever TeacherWhateverR GET
```

Don't feel strongly enough to push for that though. What you've stated here makes sense; thanks for thinking about this temporally!

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+    body <- requireJsonBody @PostTeacher
+  ```
+
+- Prefer in-module data-access functions, until they require sharing
+- Order your module as:
+  - Request/response type(s)
+  - Handler that uses it
+  - Repeat if more than one handler
+  - Internal functions
+- Document your endpoints
```
5outh, line 179: :+1: Thank you for adding this to the guide, documentation on the v2 endpoints is pretty holey.

5outh (approved):

haskell-api.md
pbrisbin: I actually don't mind that. It's nice that it requires so many fewer identifiers to build the route in Haskell

- `TeachersR`, vs
- `V2P $ TeachersP TeachersR`

Are there downsides?

I can see it'll lead to some **very** long lines in the routes files, too bad we can't do

```
/some/super/long/route/and/more/things/here/omg/80/is/coming/so/coon \
  TheThingR GET POST DELETE
```

haskell-api.md
pbrisbin: The double-negative is tripping me up, but I *think* you are asking if this guide is somehow saying that `k=v` shouldn't work? It is very much not saying that. One of the PROs of `IN` semantics listed next is that `k=v` still works as expected (like simple equality) even though you are technically saying `k IN [v]`

haskell-api.md
pbrisbin: That's right, yeah.

haskell-api.md
pbrisbin: Yup.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+- Our current `/foo/:id,:id,:id/bar` does bump into issues of length and it is
+  annoying to resolve because it means a new *route* to find a natural way to
+  group the massive list (i.e. all students in a school will mean changing
+  `/students/...` to `schools/:id`)
+- Doing the same thing in a `GET` param will be much nicer to handle when we
+  encounter it because we would probably already support both `students.id` and
+  `schools.id` in the single endpoint and the Frontend can just switch
+
+Therefore, I find the length concerns for `IN`-queries to be less problematic
```
pbrisbin: This _Aside_ was mostly to generate discussion during review. I think I'd like to remove it before merging. Anyone against that?

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+- 201 for creation
+- 204 for no content (i.e. our `Empty` responses)
+- 400s MUST follow our `ValidationError` machinery
+
+## Pagination
+
+- All list-returning routes of modest size should be paginated via `Yesod.Page`
+
+## Modules Structure
```
pbrisbin, line 101: Name tweak: _Routes and Modules Structure_

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+```
+/3/teachers
+[ { id:
+  , ...
+  , school:
+    { id:
+    , ...
+    }
+  }
+]
```
adomokos, line 43: Can you provide a sample for the request here?

```
For example for this request `/3/teachers?school.id=x` the response will be:

/3/teachers
[  { id:
    , ... 
    , school:
      { id:
      ,...
      }
  }
]
```

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+[ { id:
+  , ...
+  , schools:
+    [ { id:
+      , ...
+      }
+    ]
+  }
+]
+```
```
adomokos, line 59: I would also like to see what the request URL looks like for this response. I think it is: `/3/students?schools.id=1,2,3`.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+But here, it would be `schools.id`.
+
+If a filter exists that is not for an attribute present in the response, the
+name can be inferred by what it would look like if it were.
```
adomokos, line 64: This is super powerful, but could be confusing as well. If a request URL has multiple filters, it might be easy to miss why the response does not have the entities that I would expect. "With great power comes great responsibility." 🤷‍♂

adomokos (commented):
First of all, thank you for adding this section to the guide. It's very helpful.

It seems to me, that we are moving towards a much more powerful query mechanism, which is great, as it will reduce the number of endpoints we have to maintain, but I wonder if there is anything out there open-sourced, where we don't have to write all this ourselves. Or if we do, can we make sure that we contribute back to the community?

haskell-api.md
pbrisbin: I'll try to mention the request URL more prominently here. I only hesitate making this section look like it's trying to specify what the _response_ should look like, though, which I think your sentence starts to do. The section about how the response informs the query params, not the other way around.

pbrisbin:
@adomokos 

> I wonder if there is anything out there open-sourced, where we don't have to write all this ourselves. Or if we do, can we make sure that we contribute back to the community?

I'm curious about this. Can you give a concrete example of what kind of library, module, and/or functions you might see us building to support these ideas that would make sense to look for in, or contribute back to, the community?

haskell-api.md
pbrisbin: Heh, this is not really meant to be any kind of powerful thing... We're just talking about adding some structure around naming choices. This sentence is just to answer the obvious question about a Guide that says to "name parameters like they are in the response", which would be "well, what if it's not in the response?"

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+
+---
+
+**Aside about URL length limits**: We don't know at what size this will be a
+problem. Standards-wise, there is no limit, but browsers and servers do have
+limits and it's recommended to use 2000 as a reference point [ref].
+
+[ref]: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
+
+However, I will say this:
```
eborden: This seems like a commit message and not something that should be in the guide proper. The guide should tell you what to do, quickly and tersely. This extra exposition just clutters things up.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+  about them. For example:
+  - `students: [{id:, firstName:, lastName:}]`
+  - `school: {id:, name:}`
+
+## Status Codes
+
+- 201 for creation
+- 204 for no content (i.e. our `Empty` responses)
+- 400s MUST follow our `ValidationError` machinery
+
```
eborden, line 96:
```suggestion
- 202 for async jobs (with `Link` header)

```

haskell-api.md
eborden: There may be some issues when you have ambiguity and need to use `!`, but I haven't thought tooooo hard about it.

haskell-api.md
```diff
@@ -0,0 +1,236 @@
...
+  data PostTeacher
+
+  postTeacherR :: Handler (Entity Teacher)
+  postTeacherR = do
+    body <- requireJsonBody @PostTeacher
+  ```
+
+- Prefer in-module data-access functions, until they require sharing
+- Order your module as:
+  - Request/response type(s)
```
eborden, line 175: Does this include instances? We typically place class instances below the data type they are declaring for.

haskell-api.md
eborden: :clap:

eborden (commented):
Just a few things, I'm seconds away from approval :stuck_out_tongue_winking_eye:

haskell-api.md
pbrisbin: Yup: https://github.com/freckle/guides/pull/39#discussion_r310117292

haskell-api.md
pbrisbin: > Does this include instances? 

This document makes no claim about that one way or the other, but...

> We typically place class instances below the data type they are declaring for

I agree with this, and it should probably go in our general guide. I'd consider that a normal Haskell thing that would be followed anywhere, including `Handler`s.

eborden: Can a recommend flipping `handlers` and `Request/response types`? That order seems like the full progressive disclosure path.

pbrisbin: Hmm, I think that's too radical to do at this stage. Types first is a **strong** convention, followed generally in Haskell and very much in our own codebase. I'm a soft-no for that reason, but wouldn't block someone loudly in favor.

More importantly though, I wouldn't feel good about making that flip to get your Approve and then merging without the rest of the team re-weighing in on that specific point. Would you be willing to let this merge as-is and raising that ordering suggestion separately after?

adomokos:
> I'm curious about this. Can you give a concrete example of what kind of library, module, and/or functions you might see us building to support these ideas that would make sense to look for in, or contribute back to, the community?

I have not researched the topic, I don't know what exists out there. If there's nothing, can we build it in a way that we can extract it in the future? We usually build things that way, I don't see it as a big challenge.

haskell-api.md
eborden: Yes, I'm fine to not be too controversial.

adomokos (approved):
Thank you for adding this!

eborden (approved):
Good job whipping votes over the last few days.

haskell-api.md
jdreaver: I'm a huge proponent of progressive disclosure in module structure. I think for handler modules specifically, types first is still counts as a good form of progressive disclosure because generally the type name has the handler name in it (i.e. `TeachersGET`) and you see exactly what data will be returned up front. Then the handler function gives you details about how that is accomplished.

That's my 2 cents :man_shrugging:

eborden: I can sympathize with that reason. The problem is that practically you get:

```hs
data TheType = _

instance ...

instance ...

instance ...

instance ...

theHandler = _
```

jdreaver: That's a good point, I didn't consider the instances.

PR #41, opened 2019-07-31
Rewrite for automated approach

pbrisbin:
Much of our Guide was specifying things we now leave to tools like
Brittany. This change updates the Guide to not duplicate whatever the
current behavior of our style automation happens to be, and instead
indicates that it being automated is the real guide.

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -1,25 +1,49 @@
 # Haskell Style
 
-Principles:
+**TL;DR**: If it can be automated, it should be automated. If the automated
+outcome has no *objective* readability problems, we should accept it.
 
-* **Maximize readability**
-  * Code should be immediately readable by someone not familiar with that area.
-    If you have to continuously refer to documentation, comments or dig into
-    implementation to understand what's happening, the code isn't sufficiently
-    self-descriptive and needs to be improved. When not possible, you must
-    comment. Coding is a team sport, _always write with other developers in
-    mind_
-* **Be aware of the context**
-* **Use your judgement**
+We use three Haskell auto-formatting tools:
 
-There are some hard rules, but good style is context-sensitive. Something that
-is good in preferable in one place could be bad in another.
+- Stylish Haskell (v0.9.2.2)
+
+  We use Stylish Haskell for `import` formatting only, because it does a better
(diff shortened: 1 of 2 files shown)


adomokos:
😍

haskell-style.md
```diff
@@ -519,132 +219,38 @@ main = do
...
-module Driver (scienceOptions, socialStudiesOptions, mainWith) where
-
--- Good
-module Driver
-  ( mainWith
-  , scienceOptions
-  , socialStudiesOptions
-  ) where
-```
+Sort your exports, unless the order matters in your desired Haddock output.
```
pbrisbin, line 248: The more I think about it, the more I think I'd remove this. In actuality, exports ordering should *always* be adjusted for the best Haddocks, even in an otherwise-undocumented module. I think our Haddock-related section could imply or out-right state to do so. :man_shrugging:

adomokos:
I like that most of our formatting is automated, but seeing good/bad or just general examples is super helpful. Are we sure we want to get rid of them?

pbrisbin:
> Are we sure we want to get rid of them?

I am, because they will get out of date (some already were) and so would be super _un_-helpful (IMO) because they would be wrong. Also, part of the benefit of automation is to _never_ have to think about good or bad again. It's automatically consistent and that's it. Even a passing awareness of what that "consistent" might look like shouldn't be necessary or valuable. Therefore, the examples aren't either -- especially if they're a maintenance burden.

But that's just my (particularly staunch) take. I'm curious what the rest of the team thinks.

haskell-style.md
eborden: Agreed, the order should not be arbitrary and I consider sorting arbitrary.

eborden (approved):
LGTM.

Just one comment where I agree with your thought.

haskell-style.md
pbrisbin: So noted. I'll keep it for now and address in a separate PR.

5outh (approved):
I'm not used to seeing hugely red _guides_ PRs. ;)

cdparks (approved):

adomokos (approved):

PR #42, opened 2019-08-01
Extend Database access function naming Guide

pbrisbin:
This is just meant to canonicalize our prevailing style as far as I could tell
in a uniform way that is (relatively) easy to describe. It's not meant to
propose some "better" style that we should move everything to, but just specify
what, of the mixed styles you will see in our code, is the one we should
reproduce in new code.

### DIFF
--- a/naming-conventions.md
+++ b/naming-conventions.md
@@ -193,7 +193,7 @@ you know you need a `Foo` by its `barBat`, you should look for
 -- Good
 fetchTeachersBySchoolId :: SchoolId -> SqlPersistT m [Entity Teacher]
 
-fetchTeacherIdsBySchool :: School
+fetchTeacherIdsBySchool :: School -> SqlPersistT m [TeacherId]
 
 -- NOTE: We don't go as far as distinguishing Entity or not in this naming, so
 -- these are also Good, with Resource ~ Teacher and Resource ~ Entity School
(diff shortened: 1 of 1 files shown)


naming-conventions.md
```diff
@@ -168,18 +168,62 @@ consensus about this.
...
-```
-fetchAnswers :: SqlReadT [Answer]
+```hs
+-- Good
+fetchTeachersBySchoolId :: SchoolId -> SqlPersistT m [Entity Teacher]
+
+fetchTeacherIdsBySchool :: School
+
+-- NOTE: We don't go as far as distinguishing Entity or not in this naming, so
+-- these are also Good, with Resource ~ Teacher and Resource ~ Entity School
```
pbrisbin, line 199: I think if this were to lead to a conflict it can be resolved by:

- Upgrade return values of `e` to `Entity e` and discard the `entityKey` in callers
- Downgrade arguments from `Entity e` to `e` and discard the `entityKey` in callers

That said, I haven't seen this happen much.

eborden (approved):

cdparks (approved):

adomokos (approved):
Thank you for adding these!

5outh (approved):
:fire: We desperately needed this guide.

PR #43, opened 2019-08-01
Add StrictData guide

pbrisbin:


jdreaver (approved):

cdparks (approved):

haskell-best-practices.md
```diff
@@ -505,6 +505,11 @@ The fact is that `mono-traversable` is not an abstraction that targets our
 use-case, it decreases type inference, increases cryptic type errors, and
 increases boilerplate.
 
+## Strictness
+
+Add `LANGUAGE StrictData` to modules that contain plain records that will
```
eborden:
```suggestion
Add `LANGUAGE StrictData` to modules that contain only plain records that will
```

eborden (changes requested):

haskell-best-practices.md
```diff
@@ -505,6 +505,11 @@ The fact is that `mono-traversable` is not an abstraction that targets our
 use-case, it decreases type inference, increases cryptic type errors, and
 increases boilerplate.
 
+## Strictness
+
+Add `LANGUAGE StrictData` to modules that contain only plain records that will
```
pbrisbin, line 510: FWIW, I find this kind of risky. It might only contain plain records _right now_, and when that changes we're expecting folks to be good about removing the pragma?

eborden: If you introduce a data type that needs to be lazy, then you'll need to remove `StrictData` and switch to bang patterns. :man_shrugging:

eborden (approved):

haskell-best-practices.md
pbrisbin: "You'll need to" as in the compiler will force you?

eborden: Well the compiler won't force you to do anything, but your need for lazy semantics will.

pbrisbin: Yeah, but my point was it's risky because you'll need to remember to do so. For example, if you introduce a recursive and infinite type and forget to remove or override the pragma, you'd introduce an infinite loop right?

eborden: Yup, which is one of the things that makes me wary of `StrictData` and `Strict`, since they change the semantics of the language.

cdparks: You can either remove `StrictData` and use `!` or keep `StrictData` and use `~`

pbrisbin: But we're comfortable with requiring a human to _remember_ to do that? Perhaps a different human than the one who added the pragma to begin with? And in a PR where the diff likely won't even show the module header during review?

I think I am (hence merging this), but I think it's a valuable question to ask, and was missed in the original Issue discussion.

PR #44, opened 2019-10-10
Add explicit guide about whitespace

pbrisbin:


pbrisbin:
Love to see Restyled on a Style Guide PR :joy:, but now we have to decide if we like what `prettier-markdown` does.

haskell-style.md
```diff
@@ -41,6 +41,18 @@ done for a few reasons:
...
+## Whitespace
+
+- Prefer two space indentation
+
+  Exceptions include anything automated (e.g. de-dented `where`) and the leading
+  tabs required in `Makefile`s.
+
+- No trailing whitespace
+
+  No extra spaces at the end of lines, and no extra lines at the end of the file
```
pbrisbin, line 53: Brittany takes care of the former, but not the latter. Prettier is handling both on the JS side. There is almost certainly other language-agnostic plugins to handle it, worth finding and mentioning?

eborden: My google-fu may be busted, but I'm not finding a definitive way of doing this in `vim`.

eborden: Did you find a way to make restyled do this?

https://github.com/freckle/megarepo/pull/9793

pbrisbin: Huh, I can't seem to find a plugin that mentions that feature explicitly either. I would _guess_:

1. You're in a minority as someone who, for some reason, frequently hits enter an extra time when you are done editing the last line of a file
1. Plugins (like prettier) happen to handle this *by accident* depending on how they implement the output of the reformatted AST, but never call it out as an explicit feature, because of (1)

eborden: Oh wait no, that is from a garbo solution I was trying in `vim`.

pbrisbin: > Note that this is not "no end of line", and your vim setting above is incorrect. That would create files with no trailing newline, and that could cause actual technical problems. They will cause warnings in git, for example. You can spot the difference in GitHub too, it'll say in red No newline or some such.

That was Restyled (rather Brittany, I think), fixing that mistake.

cdparks (approved):

pbrisbin:
@eborden ping

eborden:
My blocker on this is automation. When we create a guide we expect engineers to follow it and to look for it in code review. That exacts a tax on their cognition. Without editor automation I don't see the value in having engineers manually lint whitespace that lacks semantic meaning.

haskell-style.md
```diff
@@ -41,6 +41,18 @@ done for a few reasons:
 - This gives us a chance to see just how much time we've saved by automating
   what used to be a (possibly tenuous) negotiated consensus
 
+## Whitespace
+
+- Prefer two space indentation
+
+  Exceptions include anything automated (e.g. de-dented `where`) and the leading
+  tabs required in `Makefile`s.
```
eborden, line 49: Do we have a method of automating the formatting of make files? The tabs+spaces in make files is very wacky and without automation it is just an invitation for error.

pbrisbin: > Do we have a method of automating the formatting of make files? The tabs+spaces in make files is very wacky

My editor (vim) just does it, so yes, but I don't know what's driving it -- I don't have any plugin specifically for that. Yours doesn't?

It's possible it's an Arch system configuration? Every `Makefile` I've ever come across has worked that way, but I mostly encounter `Makefile`s in an Arch or Arch-adjacent ecosystem.

pbrisbin:
> That exacts a tax on their cognition. I don't see the value in having engineers manually lint whitespace that lacks semantic meaning.

1. When anything is inconsistent, it exacts a tax on my cognition
1. Without a Guide, there will be diff noise
1. Without a Guide, there will be repeated conversations

The value is avoiding these things. It's very minor, but it's there.

> having engineers manually lint whitespace that lacks semantic meaning.

Your fingers actively add an extra newline to files, sometimes. You seem to be the only team member who does it, and you do it inconsistently. I 100% agree with you that we shouldn't impose a constant tax of manually linting whitespace (for example: manually-managed context-sensitive alignment). This is not that. This would be a very temporary and very minor breaking of a some muscle memory on your part. I'm sorry that you are the only one who bears the downside, but I think it's objectively minor when measured team-wide.

More or less minor than the value I listed above? IMO less, but it's certainly debatable.

I also rarely feel a lack of automation should block a Guide that we otherwise feel is a good thing. That's just letting Perfect be the enemy of Good.

I also do plan to automate (it's kind of Restyled's whole jam), I just don't want to block the Guide on that.

eborden (approved):
> Your fingers actively add an extra newline to files, sometimes. You seem to be the only team member who does it, and you do it inconsistently.

I think there may be some miss-understanding going on here. This comment suggests to me that you think I'm in opposition because _I am a perpetrator and don't want to change my behavior_. That is not the issue.

For there to be any understanding we may need to take it back to first principles. So what is the objective of code review? To identify elements of a patch that may be problematic in current form or in the future. As a result code review has a few different forms:

1. Design review: A 30,000 foot view that questions whether there are fundamental flaws in the overall design. Could another design option provide more beneficial effects while limiting trade-offs?
2. Bug review: We all make logical mistakes and sometimes a second set of eyes catches them. Code review can help prevent regressions and future bug reports.
3. Communication review: Does this code clearly communicate its intent? We spend more time reading code, so we should ensure it communicates clearly and effectively. Reviewers ask participants to clarify code to make it more understandable for future readers.
4. Style review: Having a consistent style can aid communication. Reviewers ask for stylistic improvements so the code will conform to the agreed upon style of the organization.

There is an essential overlap between _communication review_ and _style review_, since consistent style aids communication. However not all style has a significant impact on communication. Generally the impact is seen in areas where we have an abundance of choice. For example Haskell records:

```hs
data Foo = Foo
  { foo :: Int
  , bar :: Int
  , baz :: Int
  , biz :: Int
  , bats :: Int
  }
```
vs
```hs
data Foo = Foo { foo, bar, baz, biz, bats :: Int }
```

Neither of these forms are most stylistically right, they are choices in a stylistic space. However if one is common in a codebase and one is uncommon then it will hinder readability for most consumers. Most Haskellers don't even realize the latter is even possible. So given the abundance of choice (25 areas for variation in white space and alignment in the former) we pick one to aid communication. This style choice has a significant impact on readability.

I have a hard time believing that the two following examples have significant differences in readability.

```hs
main = putStrLn "Hello World"
```
vs
```hs
main = putStrLn "Hello World"
 
```

Yes, one may itch a desire for "cleanliness" and removal of redundancy. However the choice of whether to enforce a rule on something so minor makes a bold statement about our idea of what is _significant_ for readability. Is the litmus test for significance _redundancy_ or is it _readability_? I would hope that we are spending stylistic review time on significant stylistic errors that impede readability and therefore make code harder to maintain.

At the same time I am not happy about being the grump in the room. So since I have said my piece I will drop my block.

pbrisbin:
> At the same time I am not happy about being the grump in the room

I wouldn't classify your participation here as the grump in the room; I'm honestly trying to navigate to a good compromise, and your input has been super valuable.

Setting aside the high level philosophy for a moment, let me back up to concrete motivations and outcomes.

I'm taking the time to propose a Guide, because:

1. I would like to be able to comment on _inconsistencies_ I notice in code review
1. I would like to address _inconsistencies_ that I notice when I'm working
1. Trailing whitespace and redundant newlines are in this group of _inconsistencies_ for me

I claim no huge *significance* or *readability* gains. I don't expect the Team to suddenly take on a new level of *enforcement* in all their own Code Reviews. That's just not my litmus test for a Guide.

My litmus test is this: is it OK for someone (i.e. *me*) to have these items in their sense of consistent style without disrupting others when they collaborate within our team?

If there is a Guide:

1. I can leave a terse :art: comment, the reviewer is free to quietly ignore
1. I can feel good including these in separate "style fix" commits

If there is no Guide:

(1) could cause discussion and possible disagreement, repeatedly. (2) is now technically _unwanted_ diff noise. Alternatively, I need to remember a list of implicit things that I can't not see, but shouldn't comment on or address.

Therefore, my Guide proposal PRs are trying to uncover which of these is the case:

1. We all agree on some style, we should write it down,
1. We all disagree on some style, we should resolve that and write it down, or
1. Some of us care about and agree on some style (perhaps just one person), others don't care, but don't disagree

My micro level view is that this PR is (3) and my macro level view is that cases such as (3) should still be written down because they empower those that care with little tax on those that don't.

I think we've been discussing the macro view entirely so far, where I believe you disagree, but do we agree on my micro view?

pbrisbin:
Closing this :smile: 

1. 2-space indent is handled by Brittany anyway
1. Makefile whitespace, meh... I'll just comment in PRs and see how it goes
1. Trailing spaces and newlines, I made a Restyler over lunch

   https://github.com/restyled-io/restylers/pull/46

Thanks for the thorough discussion @eborden !

Issue #46, opened 2019-10-11
Use prettier-markdown?


pbrisbin:
[prettier-markdown](https://prettier.io/blog/2017/11/07/1.8.0.html) was recently added as a default Restyler, and since we're running default configuration in this repository, it fired on the  [next PR to open](https://github.com/freckle/guides/pull/45). 

Here are the changes it made:

- It enforces 80 columns

  We prefer and consistently do this in prose, so the only thing this affected were link references that violated:

  ```markdown
  Before
  
  [link]: http://example.com/looooooooooooooooooooooooooooooooooooooooooooong

  After
  
  [link]:
    http://example.com/looooooooooooooooooooooooooooooooooooooooooooong
  ```

- It prefers `_` over `*` for italics

  ```markdown
  Before
  
  Hi *there*

  After
  
  Hi _there_
  ```

- It prefers `-` over `*` for unordered lists

  ```markdown
  Before
  
  * Something
    * Sub-thing
  * Something else

  After
  
  - Something
    - Sub-thing
  - Something else
  ```

This is certainly not all it reformats, but it's the only things we were currently violating in the Haskell Style file, so I believe they are the salient points to discuss.

Generally, we would need to decide if we want to

1. accept what the tool does in the interest of automation,
1. configure the tool to match our preferences, or
1. disable the tool.

In this case, prettier doesn't believe in configuration, so (aside from very few [overall options](https://prettier.io/docs/en/options.html)) we can't do the second option; this is an "in or out" decision.

What do we think? Feel free to react :+1: for option (1) or :-1: for option (3).

Personally, I'm in (surprise). These choices are purely asthetic, with a very minor technical pro for prettier in that it's less overloading of `*`, which may be objectively better. The "incorrect" forms are super ingrained in my fingers, which I would have to unlearn, which means I would want to make this our general Markdown style and enable prettier-markdown on all repositories.

pbrisbin:
Piggy-back vote: if consensus is yes, should I do a big-bang reformatting in its own PR?

**EDIT**: opened and linked below.

PR #47, opened 2019-10-15
Format markdown with prettier-markdown

pbrisbin:
See #46.

I figured I would just make the big-bang and we can decide via PR, instead of Issue, if we want this or not.


eborden (approved):
automated > _artisanal_

cdparks (approved):

PR #50, opened 2020-03-10
Changes to default-extensions guide

pbrisbin:
There are three commits:

- 5fb9380 Document entities extensions caveat

- c853459 Remove caveat about NoImplicitPrelude

  We actually don't make use of this caveat. We just import Prelude explicitly,
  and it seems fine so far.

- 56a4fa1 Add DataKinds to default-extensions

  This commit has our Guide match reality, as this was already enabled in all
  packages anyway.

  DataKinds is relatively innocuous and, while we consider it an _Isolate_
  generally so as not to build things on it, we do _use_ quite a few thing that
  require it be enabled. For example:

  - Anything using a type-level symbol (Tagged, Closed, Envelope etc)
  - Any Graphula tests with type signatures (i.e. GraphulaContext m '[])

  Therefore, it makes sense in default-extensions.

I discussed this in Slack ahead of time, and don't expect much objection to
these, so I decided to just group them into one PR. If anyone wants to discuss
any one in isolation, just say so and I can slice it off separately.

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -256,6 +256,7 @@ Sort your exports, unless the order matters in your desired Haddock output.
   ```yaml
   default-extensions:
     - BangPatterns
+    - DataKinds
     - DeriveAnyClass
     - DeriveFoldable
     - DeriveFunctor
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -285,6 +285,14 @@ Sort your exports, unless the order matters in your desired Haddock output.
   extensions MUST be defined via LANGUAGE pragmas in the modules where they're
   needed.
 
+  We allow our `entities` package to diverge from this list, since it is almost
+  entirely persistent Entity definitions. Within this package only, we also have
+  the following enabled by default:
+
+  - `QuasiQuotes`
+  - `TemplateHaskell`
+  - `UndecidableInstances`
+
 - Leave a blank line after the extensions list
 
   ```haskell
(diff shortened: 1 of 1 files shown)


halogenandtoast (approved):

haskell-style.md
```diff
@@ -280,13 +281,18 @@ Sort your exports, unless the order matters in your desired Haddock output.
...
   ```
 
-  **NOTE**: `NoImplicitPrelude` may be omitted in packages using the normal,
-  implicit `Prelude` everywhere.
-
   This defines a consistent, and minimally-extended Haskell environment. Other
   extensions MUST be defined via LANGUAGE pragmas in the modules where they're
   needed.
 
+  We allow our `entities` package to diverge from this list, since it is almost
```
eborden, line 288: Kinda wish this wasn't true, but better to document than not.

eborden (approved):

haskell-style.md
pbrisbin: I'm open to making ourselves put all the pragmas in all the modules, if you wanna push on it.

TH/QQ are already explicitly in all the modules right now, but we elected to make `UndecidableInstances` special when a persistent update introduced its need. I'm purely working the consistency angle here, by moving TH/QQ in the direction of UI. If we worked the consistency angle the other way, and forced moving UI to a pragma in all the modules instead, I'd be perfectly happy.

cdparks (approved):

haskell-style.md
5outh: `entities` definitely feels special so this isn't wildly problematic for me; it makes sense and I doubt it will be questioned frequently, if at all

5outh (approved):

cbeav (approved):

PR #51, opened 2020-04-15
Update default-extensions to include DerivingVia

halogenandtoast:
Update default-extensions to include DerivingVia


eborden (approved):
:pray:

pbrisbin (commented):
Swore this was already the case. :+1:

PR #53, opened 2020-09-22
Remove avoid MonadFail

eborden:
The problem with `MonadFail` was specifically the `fail` method of
`Monad`. Because it was coupled with `Monad` is resulting in many
horrible instances, like the `Either` instance of `error` that is
mentioned here. After the Monad Fail Proposal `fail` has been pulled out
into its own class `MonadFail` that is useful and has limited instances
that aren't foot guns.

### DIFF
--- a/haskell-best-practices.md
+++ b/haskell-best-practices.md
@@ -18,54 +18,6 @@ been maintained for that as things have evolved.
       x `divEx` d = x / d
   ```
 
-### Avoid `MonadFail`
-
-Don't write functions constrained by `MonadFail`.
-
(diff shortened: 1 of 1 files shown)


pbrisbin (approved):
Works for me

5outh (approved):

jdreaver (approved):

stackptr (approved):

halogenandtoast (approved):

cbeav (approved):

cdparks (approved):

PR #58, opened 2021-03-22
Add prefer patch over put

eborden:
A recent post-mortem lead to the consensus that `PATCH` methods are
superior to `PUT`. The `PUT` method exposes many gotchas and risks in
the development process. This adds a guide to encode our beliefs.

- Post Mortem: https://renaissancelearning.atlassian.net/wiki/spaces/EN/pages/1629487213/2021-03-12+CRM+ID+Deletion+Post+Mortem
- Code Climate Error: https://github.com/freckle/megarepo/pull/15959
- Asana debt task: https://app.asana.com/0/399653121150251/1200094903267017/f

### DIFF
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -97,7 +97,7 @@ possible. Issues include:
 - The necessity of sending monolithic resources to a `PUT` when most uses are
   changing specific details.
 - The ability to misinterpret an `undefined` as a `null` and have untintended
-  consequences when evolving a handler. This is exasterbated by the default
+  consequences when evolving a handler. This is exacerbated by the default
   parsing behavior of `Maybe a` in `aeson`.
 - The complexities that arise from extra validation or lack of validation
   especially when differing rules around mutability and roles arise.
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -87,6 +87,24 @@ This is because:
   - `students: [{id:, firstName:, lastName:}]`
   - `school: {id:, name:}`
 
+## HTTP Methods
+
+### Prefer `PATCH` over `PUT`
+
+`PUT` routes contain semantics that easily lead to bugs. We prefer `PATCH` when
+possible. Issues include:
+
+* The necessity of sending monolithic resources to a `PUT` when most uses are
+  changing specific details.
+* The ability to misinterpret an `undefined` as a `null` and have untintended
+  consequences when evolving a handler. This is exasterbated by the default
+  parsing behavior of `Maybe a` in `aeson`.
+* The complexities that arise from extra validation or lack of validation
+  especially when differing rules around mutability and roles arise.
+
+If you are creating a new `PUT` consider if it could be expressed in a more
+granular `PATCH` semantic.
+
 ## Status Codes
 
 - 201 for creation
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -94,12 +94,12 @@ This is because:
 `PUT` routes contain semantics that easily lead to bugs. We prefer `PATCH` when
 possible. Issues include:
 
-* The necessity of sending monolithic resources to a `PUT` when most uses are
+- The necessity of sending monolithic resources to a `PUT` when most uses are
   changing specific details.
-* The ability to misinterpret an `undefined` as a `null` and have untintended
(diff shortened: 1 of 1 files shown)


joris974:
Could we create a task to identify all PUT as debt and convert them to PATCH?

eborden:
@joris974 see the linked task above.

pbrisbin (approved):

StevenXL (approved):

stackptr (approved):
I wonder if this repo could be made more visible somehow. My first impulse to find this would be to go to Confluence

mjgpy3 (approved):
I [certainly agree](https://freckleinc.slack.com/archives/C8FUUKHHC/p1611355416011000?thread_ts=1611352897.007900&cid=C8FUUKHHC)!

luckysoni (approved):
Slightly related: Should/Does an entity maintain a whitelist of properties that can be mass assigned? https://laravel.com/docs/8.x/eloquent#mass-assignment

cdparks (approved):

cbeav (approved):

cristygrant (approved):

eborden:
Looks like this is a quorum. Merging.

@luckysoni I'm not sure I totally understand the mass assignment topic in that Eloquent doc. As far as I'm aware this is a peculiarity of the way that framework decides to hand HTTP requests. It seems it dynamically will accept any parameters passed to an HTTP request. We specify and parse all our HTTP requests, so we aren't necessarily open to this vector.

> A mass assignment vulnerability occurs when a user passes an unexpected HTTP request field and that field changes a column in your database that you did not expect. For example, a malicious user might send an is_admin parameter through an HTTP request, which is then passed to your model's create method, allowing the user to escalate themselves to an administrator.

luckysoni:
> We specify and parse all our HTTP requests, so we aren't necessarily open to this vector.

That is good to know @eborden. 

So if the PATCH request body is an object which tries to update the `id` (or any other sensitive column) of an entity, do we disallow that operation? Do we have a list of updatable columns for each endpoint? Just curious how we address mass assignment.

pbrisbin:
> do we disallow that operation? Do we have a list of updatable columns for each endpoint? Just curious how we address mass assignment

This sort of thing is common in a dynamic language where the patch body is an un-typed `Object` and it just gets passed to `Model.update(body)`. Nowadays, even a framework like Rails is set up to not allow that: you can't just pass things in like that without explicitly validating things (i.e. `StrongParameters`).

But in Haskell it's just impossible to get even close to that. Every endpoint has a type that represents is body and only contains parameters that endpoint expects:

```hs
data PatchSomething = PatchSomething
  { name :: Maybe Text
  , grade :: Maybe Grade
  }
  deriving stock Generic
  deriving anyclass FromJSON

patchSomethignR :: SomethingId -> Handler Empty
patchSomethingR somethingId = do
  body <- requireJsonBody @PatchSomething
```

Passing any keys other than `name` or `grade` will be a 400 failure to parse the JSON.

You could choose to not have a dedicated body type and just parse a generic `Value` (Haskell's type for any JSON Object), but that type can't just be given to an ORM update function directly since that wouldn't type-check either.

```hs
  -- Using the request body type:
  update somethingId $ catMaybes
    [ (SomethingName .=) <$> name body
    , (SomethingGrade =.) <$> grade body
    ]
  pure Empty

  -- Using Value and generic lookup functions
  update somethingId $ catMaybes
    [ (SomethingName .=) <$> (body ^? key "name" . _JSON)
    , (SomethingGrade =.) <$> (body ^? key "grade" . _JSON)
    ]
  pure Empty
```

The specifics of this aren't important here; I'm just trying to illustrate that it's all very, very explicit where things come from and where they go. So tossing `id: x`, or `admin: true` in a `PATCH` body wouldn't get you very far. It's possible, but it would take a lot of other mistakes along the way.

pbrisbin:
Quick correction:

> Passing any keys other than name or grade will be a 400 failure to parse the JSON.

This isn't true. We generally ignore unknown keys, so extra keys would not be a 400. But since we can only access data through the parsed type, they couldn't cause us any problems either.

luckysoni:
Nice! Thank you for the context.

PR #63, opened 2021-07-20
Link to long running operations from 202 guide

### LINKED ISSUE #57: Link vs Location
In [Status Codes](https://github.com/freckle/guides/blob/master/haskell-api.md#status-codes), should "202 for async jobs (with **Link** header)" read "202 for async jobs (with **Location** header)"?

pbrisbin:
Closes #57.

### DIFF
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -108,9 +108,11 @@ granular `PATCH` semantic.
 ## Status Codes
 
 - 201 for creation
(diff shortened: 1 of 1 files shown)


pbrisbin:
Ignoring Restyled. It's a bug that pretty-markdown tries to re-indent long anchors; that actually breaks in some renderers.

StevenXL (approved):

PR #65, opened 2021-09-15
Document a bit more about query parameters

pbrisbin:
Please see individual commits.

### DIFF
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -30,7 +30,21 @@ Bad:
 - `/3/teachers/:id/students`
 - `/3/schools/:id/teachers/:id/students`
 
+### Operations
 
+For complex filtering, follow an "operation suffix" syntax,
+
+Good:
+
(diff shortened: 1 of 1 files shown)


haskell-api.md
```diff
@@ -75,20 +89,6 @@ But here, it would be `schools.id`: `/3/students?schools.id=1,2,3`.
...
 name can be inferred by what it would look like if it were.
 
-### Semantics
-
-Prefer `IN` equality semantics.
-
-- Good: `{attribute}={value1},{value2},{value3},...`
-- Bad: `{attribute}={value}`
-
-This is because:
```
pbrisbin, line 73: I didn't think this justification was super valuable, so once I mentioned the "behave like `[in]`" in the new section, I figured we could drop all this. If anyon disagrees, I can move this content up there instead.

eborden (approved):

PR #67, opened 2021-11-10
Add `PATCH` semantics

mjgpy3:
Having recently converted our remaining `PUT` routes to `PATCH` I thought it would be nice to unify and formalize some of the differing practices in the API.

I believe that some of these formalizations could benefit our backend testing and frontend implementations.

If `PATCH` returns the same value as `GET`, and is the same route, then frontend
 - wouldn't need to re-fetch updated entities (not sure if we do this often, I've seen it a good deal in past lives)
 - could reuse types and code used by the original `GET`
 - would (theoretically) need to manage less state since it could just trust what `PATCH` yields rather than having it's own duplicate state store to maintain
 
These formalizations are nice for the backend because
 - we should be able to generate tests from a description of the fields a `PATCH` expects and expected outcomes thereabouts
 - allowing for empty `PATCH` objects and re-fetching would allow us to implement `GET /some/route/123` as `PATCH {} /some/route/123`, or, conversely, we could implement the re-fetching using the `GET` (though that's not really a benefit).
 
Note that  our current `PATCH` endpoints do not conform to these guidelines, in many cases. If this guideline is accepted I will file debt for the places that don't conform for the backend guild to prioritize.

### DIFF
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -126,8 +126,9 @@ Modifications to an existing resource should be made via a `PATCH` request.
 A well-formed `PATCH` endpoint
 
 - Must update an entity's `updatedAt` field, if present.
-- Should return the modified resource. The resource should be re-fetched for
-  ease of testing and to ensure the update was correctly persisted.
+- Should return the modified resource. The resource should be re-fetched from
+  the database for ease of testing and to ensure the update was correctly
+  persisted.
 - Must accept `null` to unset nullable fields, disallowing it for
   non-nullables.
 - Must 404 if a resource with the route's identifier(s) does not exist.
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -182,7 +182,7 @@ For example, assuming `GET /3/teachers/7654` yields
   "phoneNumber": null
 }
 
-// Do nothing, no-op update
+// Do nothing, still updates `updatedAt`
 {
 }
 ```
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -117,6 +117,102 @@ possible. Issues include:
 If you are creating a new `PUT` consider if it could be expressed in a more
 granular `PATCH` semantic.
 
+### `PATCH` semantics
+
+Modifications to an existing resource should be made via a `PATCH` request.
+
+### Specific guidelines
+
+A well-formed `PATCH` endpoint
+ - Should update an entity's `updatedAt` field, if present.
+ - Should return an empty response or the modified resource. Ideally, if the
+   latter, the resource should be re-fetched for ease of testing and to ensure
+   the update was correctly persisted.
+ - Should accept `null` to unset nullable fields, disallowing it for
+   non-nullables.
+ - Should 404 if a resource with the route's identifier(s) does not exist.
+ - Should not allow for update-scoping information within the `PATCH` object. If
+   scoping information is absolutely necessary, it should be include in query
+   params.
+ - Should not create resources (that's the role of `POST`).
+ - Should allow for fields to be optional, even allowing for entirely empty
+   `PATCH` objects.
+ - Should not allow for unsupported keys.
+
+#### Structural similarities to `GET`
+
+As a rule-of-thumb for a given resource, `PATCH` requests share a similar shape
+to `GET` requests in that
+ - they should have the same route (e.g. `/3/teachers/7654` would `GET` or
+   `PATCH` the teacher with id `7654`),
+ - `PATCH` objects should resemble the fields yielded by the corresponding `GET`,
+   albeit `PATCH` fields are (1) optional, (2) may allow additional fields or (3)
+   disallow the modification of certain fields (e.g. `id`, `createdAt`), and
+ - if a `PATCH` response is non-empty, it should be the same payload that would
+   be returned by a call to the corresponding `GET`.
+
+#### Example
+   
+For example, assuming `GET /3/teachers/7654` yields
+```json
+{
+  "id": 7654,
+  "givenName": "John",
+  "surname": "Kimble",
+  "email": "jk@example.com",
+  "phoneNumber": "555-555 5555",
+  "addressLines": ["1234 Hollywood Dr., Hollywood, CA"],
+  "administrativeArea": "CA",
+  "country": "USA",
+  "gradesTaught": ["K"],
+  "createdAt": "2021-11-10T15:29:16.239Z",
+  "updatedAt": "2021-11-10T15:29:16.239Z",
+}
+```
+
+`PATCH /3/teachers/7654` could accept the following update payloads
+``` javascript
+// Change the teacher's name
+{
+  "givenName": "Arnold",
+  "surname": "Schwarzenegger"
+}
+
+// Remove the teacher's phone number
+{
+  "phoneNumber": null
+}
+
+// Do nothing, no-op update
+{
+}
+```
+
+However, `PATCH /3/teachers/7654` would fail given the following payloads
+``` javascript
+// BAD REQUEST, trying to unset a required field
+{
+  "email": null
+}
+
+// BAD REQUEST, trying to set a field that cannot be modified
+{
+  "createdAt": "2019-11-10T15:29:16.239Z"
+}
+
+// BAD REQUEST, trying to set an unsupported field
+{
+  "unsupportedField": "hi"
+}
+
+// BAD REQUEST, validation found `ZZ` is not within `USA`
+{
+  "addressLines": ["1234 Hollywood Dr., Hollywood, ZZ"],
+  "administrativeArea": "ZZ",
+  "country": "USA"
+}
+```
+
 ## Status Codes
 
 - 201 for creation
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -126,9 +126,8 @@ Modifications to an existing resource should be made via a `PATCH` request.
 A well-formed `PATCH` endpoint
 
 - Must update an entity's `updatedAt` field, if present.
-- Must return an empty response or the modified resource. If the latter, the
-  the resource should be re-fetched for ease of testing and to ensure
-  the update was correctly persisted.
+- Should return the modified resource. The resource should be re-fetched for
+  ease of testing and to ensure the update was correctly persisted.
 - Must accept `null` to unset nullable fields, disallowing it for
   non-nullables.
 - Must 404 if a resource with the route's identifier(s) does not exist.
--- a/haskell-api.md
+++ b/haskell-api.md
@@ -124,36 +124,39 @@ Modifications to an existing resource should be made via a `PATCH` request.
 ### Specific guidelines
 
 A well-formed `PATCH` endpoint
- - Should update an entity's `updatedAt` field, if present.
- - Should return an empty response or the modified resource. Ideally, if the
-   latter, the resource should be re-fetched for ease of testing and to ensure
-   the update was correctly persisted.
- - Should accept `null` to unset nullable fields, disallowing it for
-   non-nullables.
- - Should 404 if a resource with the route's identifier(s) does not exist.
- - Should not allow for update-scoping information within the `PATCH` object. If
-   scoping information is absolutely necessary, it should be include in query
-   params.
- - Should not create resources (that's the role of `POST`).
- - Should allow for fields to be optional, even allowing for entirely empty
-   `PATCH` objects.
- - Should not allow for unsupported keys.
+
+- Should update an entity's `updatedAt` field, if present.
+- Should return an empty response or the modified resource. Ideally, if the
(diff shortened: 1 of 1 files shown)


haskell-api.md
```diff
@@ -117,6 +117,107 @@ possible. Issues include:
...
+- Should accept `null` to unset nullable fields, disallowing it for
+  non-nullables.
+- Should 404 if a resource with the route's identifier(s) does not exist.
+- Should not allow for update-scoping information within the `PATCH` object. If
+  scoping information is absolutely necessary, it should be include in query
+  params.
+- Should not create resources (that's the role of `POST`).
+- Should allow for fields to be optional, even allowing for entirely empty
+  `PATCH` objects.
+- Should not allow for unsupported keys.
```
pbrisbin: I think we should remove this point and bring it in by a separate PR with isolated discussion.

Aeson doesn't make this easy and we do it nowhere (that I know of). If we did do it, it would decrease flexibility in backwards-compatible changes. Client couldn't be updated to send new fields first, before Server is updated to handle it. This is (probably) always solvable, but it can be non-ideal, for example, if Server changes are more complex so you'd rather maintain compatibility by doing Client first and make a single change from old -> new in the Server.

haskell-api.md
```diff
@@ -117,6 +117,107 @@ possible. Issues include:
...
+- Should update an entity's `updatedAt` field, if present.
+- Should return an empty response or the modified resource. Ideally, if the
+  latter, the resource should be re-fetched for ease of testing and to ensure
+  the update was correctly persisted.
+- Should accept `null` to unset nullable fields, disallowing it for
+  non-nullables.
+- Should 404 if a resource with the route's identifier(s) does not exist.
+- Should not allow for update-scoping information within the `PATCH` object. If
+  scoping information is absolutely necessary, it should be include in query
+  params.
```
pbrisbin: I'm not sure what this means, maybe a `Good/Bad` code listing would be useful?

haskell-api.md
```diff
@@ -117,6 +117,107 @@ possible. Issues include:
...
+### `PATCH` semantics
+
+Modifications to an existing resource should be made via a `PATCH` request.
+
+### Specific guidelines
+
+A well-formed `PATCH` endpoint
+
+- Should update an entity's `updatedAt` field, if present.
+- Should return an empty response or the modified resource. Ideally, if the
```
pbrisbin: This this is a Should anyway, why not just say it should return the modified resource? If we have cases that must return empty, consider those taking advantage of this as a should-not-must. Also, returning **modified** resource, implies re-fetching to me, so I would just cut the second sentence myself.

```suggestion
- Should return the modified resource.
```


That said, should any of these be upgraded to a MUST?

haskell-api.md
mjgpy3: Ah, yes, right! Let me remove that.

haskell-api.md
mjgpy3: Yeah, this one is tricky to talk about!

It's summarizing a pattern I undid in our code where we had a request like
```
PUT /2/courses/1234/enabledDashboardModes
{
  modes: [...],
  grade: null|number
}
```

`grade` would scope the request to only modify for certain grades. So this would be suggesting
```
PATCH /2/courses/1234/enabledDashboardModes?[grade=number]
{
  modes: [...]
}
```
instead.

Does that make sense? Perhaps this is not worth mentioning.

haskell-api.md
mjgpy3: That's a good point! I preferred "should language" because I have trouble reconciling "guidelines" with "must" but I agree.

haskell-api.md
pbrisbin: I agree that is an odd behavior, and would be good to avoid. I'm good to keep it as long as there's an example shown. I'm also fine if you want to leave it off for now.

haskell-api.md
mjgpy3: @pbrisbin how's that look?

haskell-api.md
```diff
@@ -117,6 +117,98 @@ possible. Issues include:
...
+### `PATCH` semantics
+
+Modifications to an existing resource should be made via a `PATCH` request.
+
+### Specific guidelines
+
+A well-formed `PATCH` endpoint
+
+- Must update an entity's `updatedAt` field, if present.
+- Must return an empty response or the modified resource. If the latter, the
```
pbrisbin: What is the argument for empty response being allowed?

pbrisbin (approved):

haskell-api.md
mjgpy3: Personally I'm in favor removing the `empty` affordance and just saying we must return the modified resource. But if I had to contrive a few reasons for allowing empty they'd be
 - easier/simpler implementation
 - forces consumer to re-fetch or maintain state, so it avoids the "whoops my patch returned an inconsistent object" (e.g. made updates in-memory and forgot to persist)
 - potentially more performant implementation (no refetch is necessary)

pbrisbin: Gotcha. Given that we only have contrived reasons, it feels more appropriate to downgrade this to SHOULD without the `empty` affordance. That way, cases that do need it won't be violating a MUST, and we can discover them and update the guide if necessary based on concrete cases. The way it's written now, an empty response is a first-class option that some may choose without any nudge the other way -- which, I think, is not what we'd want ideally-speaking. Right?

That said, I Approved along with this question so I'm also fine keeping it as-is.

mjgpy3: Should w/o `empty` is good with me!

haskell-api.md
```diff
@@ -117,6 +117,97 @@ possible. Issues include:
...
+
+- Must update an entity's `updatedAt` field, if present.
+- Should return the modified resource. The resource should be re-fetched for
+  ease of testing and to ensure the update was correctly persisted.
+- Must accept `null` to unset nullable fields, disallowing it for
+  non-nullables.
+- Must 404 if a resource with the route's identifier(s) does not exist.
+- Must not create resources (that's the role of `POST`).
+- Must allow for fields to be optional, even allowing for entirely empty `PATCH`
+  objects.
```
eborden, line 137: Should this be a validation failure? We have a number of `PATCH` endpoints that validate whether the payload is not `{}` and throw if it is. That leads to a lot of noisy checking, but the supposed benefit is that it catches spurious API calls and logic errors. On the other hand validating `not {}` breaks robustness principles.

eborden: random thought: If you allow empty, does `PATCH /x {}` become a synonym for `GET /x`?

haskell-api.md
```diff
@@ -117,6 +117,97 @@ possible. Issues include:
...
+}
+
+// Remove the teacher's phone number
+{
+  "phoneNumber": null
+}
+
+// Do nothing, no-op update
+{
+}
```
eborden: Is this a no-op or is it a `touch`? Does this impact the `updatedAt` of the resource?

haskell-api.md
```diff
@@ -117,6 +117,97 @@ possible. Issues include:
...
+```javascript
+// BAD REQUEST, trying to unset a required field
+{
+  "email": null
+}
+
+// BAD REQUEST, trying to set a field that cannot be modified
+{
+  "createdAt": "2019-11-10T15:29:16.239Z"
+}
```
eborden: Should this fail or should it ignore the value? Robustness principles would say that extra JSON properties should be accepted and ignored.

haskell-api.md
pbrisbin: > Should this be a validation failure? We have a number of PATCH endpoints that validate whether the payload is not {} and throw if it is.

Gilli and I discussing that our present "no empty patch" validation is unnecessary is what prompted this Guide.

> On the other hand validating not {} breaks robustness principles

That is where I/we landed, hence this guide making the push that way.

haskell-api.md
pbrisbin: Good question!

eborden: If you want this to be a true no-op, then you get all the noisy crap that comes along with validating `not {}`

haskell-api.md
pbrisbin: > Robustness principles would say that extra JSON properties should be accepted and ignored

I agree with this as stated, but I think Gilli is drawing a distinction between this and attempting to mutate a known-but-immutable field.

I can get behind ignoring either since it feels reasonable to extend that robustness principle to known-but-immutable and it will result in less implementation by anyone following the Guide.

haskell-api.md
pbrisbin: I vote it implies "anything that happens for all updates still happens" (`updatedAt` is an example).

It's a simpler implementation, and we actually have a concrete use-case ourselves: we use `PATCH /sessions/:id {}` to mark a session as started.

haskell-api.md
mjgpy3: > random thought: If you allow empty, does PATCH /x {} become a synonym for GET /x

@eborden yep, I tried to note that in the PR description. Though I guess seeing the other discussion around `updatedAt` is potentially making this not so.

haskell-api.md
mjgpy3: I fixed the comment to specify `updateAt` is updated.

haskell-api.md
mjgpy3: Clarified this to say "immutable fields are ignored" same as unsupported.

haskell-api.md
```diff
@@ -117,6 +117,97 @@ possible. Issues include:
...
+### Specific guidelines
+
+A well-formed `PATCH` endpoint
+
+- Must update an entity's `updatedAt` field, if present.
+- Should return the modified resource. The resource should be re-fetched for
+  ease of testing and to ensure the update was correctly persisted.
+- Must accept `null` to unset nullable fields, disallowing it for
+  non-nullables.
+- Must 404 if a resource with the route's identifier(s) does not exist.
```
stackptr, line 134: Should there be a call out "Must not 403..." to reflect the effort of standardizing endpoint rejections?

haskell-api.md
```diff
@@ -117,6 +117,97 @@ possible. Issues include:
...
+### `PATCH` semantics
+
+Modifications to an existing resource should be made via a `PATCH` request.
+
+### Specific guidelines
+
+A well-formed `PATCH` endpoint
+
+- Must update an entity's `updatedAt` field, if present.
+- Should return the modified resource. The resource should be re-fetched for
```
stackptr: "re-fetched" is a bit confusing to me as I assume you're talking about refreshing from the database and not refetching from a different endpoint. Could just be me, but we have `refreshEntity` and `refreshRecord` in the Persistent module so maybe this should use similar verbiage

haskell-api.md
```diff
@@ -117,6 +117,97 @@ possible. Issues include:
...
+// Remove the teacher's phone number
+{
+  "phoneNumber": null
+}
+
+// Updates `updatedAt`
+{
+}
+
+// Updates `updatedAt` (immutable fields and unsupported fields are ignored)
```
stackptr, line 190: I agree that unsupported fields should be ignored, but it feels to me that the frontend attempting to set an immutable field is very similar as trying to _unset_ a required field, because in both instances there is a disconnect between what is possible to do with the endpoint and what the endpoint allows. What is the difference that makes one of these benign/ignored and the other an error?

haskell-api.md
mjgpy3: I think that we should add a separate guide entry for 403s in general and defer to that.

haskell-api.md
mjgpy3: Deferring to Pat's comment here for this one: https://github.com/freckle/guides/pull/67#discussion_r746763879

haskell-api.md
mjgpy3: Clarified in recent commit, thanks!

haskell-api.md
stackptr: That comment is for unsupported fields. Does it apply here because immutable fields require similar implementation hurdles?

mjgpy3: Yes, I believe so. Making immutable fields fail would require us to have a first-class representation for them.

Also, if we made some immutable field mutable at a future date our API behavior would change (e.g. stop returning 400s). We'd never do this with `createdAt` but we may with other things (e.g. teacher's validated status was conceived as a computed field that may eventually become a persistent field).

eborden (approved):
LGTM. Looking forward to the discussion on immutable fields :smile:

stackptr (approved):

joris974 (approved):
Is there a way to manually or automatically detect which exisitng endpoints are not following these rules? Is there a way to check these rules automatically when creating new endpoints?

StevenXL (approved):

mjgpy3:
@joris974 I think and hope so!

PR #74, opened 2022-06-07
Note `haskell-library-template` in OSS instructions

stackptr:
This was added recently and provides a good starting point for Haskell libraries.

### DIFF
--- a/haskell-open-source.md
+++ b/haskell-open-source.md
@@ -2,8 +2,8 @@
 
 1. If authoring a distinct package of reasonable size and generic utility,
    prefer starting a separate, public repository utilized as a Git dependency.
-   A Haskell library template includes common setup for tooling and actions:
-   `gh repo create --template freckle/haskell-library-template`
+
+Templates may exist for common cases, such as a [Haskell library][gh-hlt].
 
 When working in a public repository:
 
@@ -66,3 +66,5 @@ To determine what changes are "breaking", use the following:
 ![](./pvp-chart.png)
 
 With `A=EPOCH`, `B=MAJOR`, and `C=MINOR`.
+
+[gh-hlt]: https://github.com/freckle/haskell-library-template
--- a/haskell-open-source.md
+++ b/haskell-open-source.md
@@ -1,7 +1,9 @@
 # Best Practices
 
 1. If authoring a distinct package of reasonable size and generic utility,
-   prefer starting a separate, public repository utilized as a Git dependency
+   prefer starting a separate, public repository utilized as a Git dependency.
+   A Haskell library template includes common setup for tooling and actions:
+   `gh repo create --template freckle/haskell-library-template`
 
 When working in a public repository:
 


haskell-open-source.md
```diff
@@ -1,7 +1,9 @@
 # Best Practices
 
 1. If authoring a distinct package of reasonable size and generic utility,
-   prefer starting a separate, public repository utilized as a Git dependency
+   prefer starting a separate, public repository utilized as a Git dependency.
+   A Haskell library template includes common setup for tooling and actions:
+   `gh repo create --template freckle/haskell-library-template`
```
pbrisbin: Totally minor and not blocking, but to me this fees like distinct from the best practice itself. I'm not sure if we're saying using the template is a best-practice or just an option. I might separate it from the main line-item,

> 1. If authoring... as a Git dependency
>
>    Templates may exist for common cases, such as a [haskell library](https://github.com/freckle/haskell-library-template).

I'm not sure if putting `gh repo` here is necessary either -- that feels more like usage example content, better served by the template's README or other documentation.

pbrisbin (approved):

haskell-open-source.md
stackptr: I think if this was an article for general OSS libraries I'd agree on a separate line-item, but this is specifically for Haskell OSS so I'm not sure what other common cases we'd be accounting for here. I agree that the command invocation is better left in the repo itself

stackptr: Although looking at the rest of this page, maybe this is a generic OSS page that could be renamed to reflect that?

pbrisbin: > I'd agree on a separate line-item

Just to clarify, I wasn't suggesting a "separate line-item" as in a new numbered point. I was suggesting pulling it out of the main "title" of the line-item and into the nested content as a sort of extra, because:

> I'm not sure if we're saying using the template is a best-practice or just an option

And I think it's the latter.

> I'm not sure what other common cases we'd be accounting for here

I don't totally follow this. Not sure why or what "other common cases" would determine where/how this point should be included.

> maybe this is a generic OSS page that could be renamed to reflect that?

The versioning content is Haskell-specific, the rest is not.

We're in a weird place where we have this content here and other content [in confluence](https://renaissancelearning.atlassian.net/wiki/spaces/EN/pages/1442513923/Maintaining+Haskell+Open+Source) (which I need to update to mention the template too), so figuring out what should live where and what is Haskell-specific vs general OSS would be great.

stackptr: > I don't totally follow this. Not sure why or what "other common cases" would determine where/how this point should be included.

The filename being `haskell-open-source.md` made me think there would only be one case in creating a library from this page, which is to use the template—or not; but my point is that there are unlikely to be different templates to choose from. If this was just `open-source.md` I would imagine a future where there could be a template for a Haskell library, JavaScript library, etc. Then the wording would be less prescriptive.

pbrisbin: > but my point is that there are unlikely to be different templates to choose from

Well, regardless of the rest, this is not correct: we named it `haskell-library-template` because we assume `haskell-cli-template`, `haskell-yesod-template`, etc are all future possibilities.

> I would imagine a future where there could be a template for a Haskell library, JavaScript library

There will be (are) `-template` repos for lots of things beyond Haskell, so that part's true.

PR #76, opened 2022-07-01
Add a section on Testing patterns

pbrisbin:


haskell-best-practices.md
```diff
@@ -473,6 +473,214 @@ increases boilerplate.
...
+  ) where
+```
+
+### Imports
+
+Each project should have a `TestImport` used for all specs. Optionally, a
+`TestImport.IO` version can exist for writing non-App specs in `IO`.
+
+```hs
+-- Bad
```
mjgpy3: Please add comments pointing to exactly what's bad to make it obvious for me, the dumb reader.

haskell-best-practices.md
```diff
@@ -473,6 +473,214 @@ increases boilerplate.
...
+shouldApproximate :: Double -> Double -> Expectation
+
+-- Good
+shouldApproximate :: HasCallStack => Double -> Double -> Expectation
+
+-- Best (see below)
+shouldApproximate
+  :: (HasCallStack, MonadIO m)
```
mjgpy3, line 562: Nice suggestion

haskell-best-practices.md
```diff
@@ -473,6 +473,214 @@ increases boilerplate.
...
+```hs
+-- Bad
+assertEqual "..." a b
+
+-- Good
+a `shouldBe` b
+```
+
+Except for functions that return values as part of their assertions, like
+`assertJust`, or `assertRight`.
```
mjgpy3, line 591: Should we settle on the use of these things vs e.g. `Just thingy <- ...`?

mjgpy3: We also have like `shouldBeJust` floating around. What should we do with those?

haskell-best-practices.md
```diff
@@ -473,6 +473,214 @@ increases boilerplate.
...
+`MonadFail(fail)` as `expectationFailure`.
+
+For example, a test about a user in the DB can incorporate that assertion,
+
+```hs
+-- Meh
+user <- assertJust =<< getDB userId
+userFoo user `shouldBe` Foo
+
+-- Nice
```
mjgpy3, line 624: Cool! Thanks for mentioning. Should we prefer one pattern or another?

haskell-best-practices.md
```diff
@@ -473,6 +473,214 @@ increases boilerplate.
...
+Just user <- getDB userId
+userFoo user `shouldBe` Foo
+```
+
+#### Example Descriptions
+
+Somehow the "should" Railsism took hold in our code-base. There's no need for it
```
mjgpy3, line 631: :raised_hands: love it, death to "should", long live "should"

haskell-best-practices.md
```diff
@@ -473,6 +473,214 @@ increases boilerplate.
...
+
+Also, avoid the word "successfully". When expecting something to happen, that
+you expect it to happen "successfully" can be assumed.
+
+```hs
+-- Bad
+it "should successfully deduplicate the users"
+  $ withGraph
+  $ do
+    omgI'mAllTheWayOverHereNow
```
mjgpy3, line 642: LOL

haskell-best-practices.md
```diff
@@ -473,6 +473,214 @@ increases boilerplate.
...
+- Using `prop` under `withApp`
+
+  Pull it out.
+
+- Using `shouldX` with `prop`
+
+  Use a pure `Bool` expression with `prop`.
+
+If you do run into ambiguity, prefer Type Applications (over type signatures or
+annotations) to resolve it.
```
mjgpy3: Do you mind linking to `TypeApplications` docs? I didn't know about this extension prior to starting and many folks I've interviewed lately don't know about it eiter.

mjgpy3 (approved):
LGTM, some comments take 'em or leave 'em. Great to see the specs getting some love.

haskell-best-practices.md
pbrisbin: `assertJust` and `Just x <-` are the same in that you get the `x`, but `shouldBeJust` is different as it's an expectation and doesn't return the `x`: ``shouldBeJust = (`shouldSatisfy` isJust) :: m ()``.

(If I'm wrong and `shouldBeJust` actually exists in a form that _does_ return the `x`, well then it should be renamed/replaced by `assertJust`.)

As for `assertJust` vs a pattern-match fail, I'm always down to pick a One True Way, but the usage feels about 50/50 in our code-bases. Dare you take a stand?

haskell-best-practices.md
pbrisbin: Answered in above thread, I think.

pbrisbin:
@mjgpy3 check it now

PR #82, opened 2023-05-19
Update docs to reference Fourmolu.

StevenXL:
Updating docs per these directions:

https://renaissancelearning.atlassian.net/wiki/spaces/ARCH/pages/42798940161/Style+Haskell+with+Fourmolu

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Best practices in our Haskell API specifically.
 
 ### [Style](./haskell-style.md)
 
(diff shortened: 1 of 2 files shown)


pbrisbin (approved):
Nice!

At one point I made a push to make sure the actual Haskell _code_ shown throughout our guides was conformant (e.g. Brittany-styled). If there are examples where Fourmolu does something differently, we'll want to update those. No need for that in this PR, just FYI.

PR #86, opened 2023-10-13
Add guides on use of unsafe functions from `Prelude`

pbrisbin:
This is a pattern I've been doing and I think it's useful. I'd like to
make it a team-wide guide.

### DIFF
--- a/haskell-best-practices.md
+++ b/haskell-best-practices.md
@@ -55,6 +55,8 @@ been maintained for that as things have evolved.
   import Yesod.Auth.OAuth2.Prelude -- Library with custom prelude module
   ```
 
+- Local preludes that re-export `Prelude` should hide unsafe functions
+
 - Libraries must use `Prelude`. Do not force an alternative prelude choice on
   end-users.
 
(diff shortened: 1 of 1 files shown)


pbrisbin:
Added a bunch of reviewers for visibility. I'll merge on a few approves, assuming no dissents.

z0isch (approved):
Never do anything unsafe and we don't need this!

joelmccracken (approved):

joelmccracken:
alternative, perhaps worth bringing up: don't ever import them, but instead as a standard use pattern matching and `error`, which gives you a good spot to write _why_ its impossible/why the partial function is acceptable in this case. Thoughts? The best we can get in haskell for now is an informal proof of correctness etc.

joelmccracken:
also, I do use partial functions in test on my personal stuff all the time; i think they're appropriate there. Downside is when it fails often the errors are bad, but it can make things so much easier/worth testing.

chris-martin (approved):

PR #91, opened 2024-05-24
Add 'TypeOperators' to default-extensions.

StevenXL:
The 'TypeOperators' extension is documented here:

- https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_operators.html#extension-TypeOperators

It is included in 'GHC2021', and thus enabling this extension is consistent with this guide's current recommendation to use 'GHC2021'.

Furthermore, the use of the type equality type operator - '~' - **without** the use of the 'TypeOperators' extension will become an error in future GHC releases.

**If** this guide suggestion is approved, there is a PR to adhere to the recommendation in megarepo:

https://github.com/freckle/megarepo/pull/34415

### DIFF
--- a/haskell-style.md
+++ b/haskell-style.md
@@ -255,6 +255,7 @@ Sort your exports, unless the order matters in your desired Haddock output.
     - OverloadedStrings
     - QuasiQuotes
     - TypeFamilies
+    - TypeOperators
   ```
 
   This defines a consistent, and minimally-extended Haskell environment. Other


tomcarste (approved):

OlaoluwaM (approved):

jason-lieb (approved):

pbrisbin (commented):
> It is included in 'GHC2021', and thus enabling this extension is consistent with this guide's current recommendation to use 'GHC2021'.

I don't get this. If it's already included in `GHC2021`, then this guide's recommendation to use `GHC2021` means it should **not** be enabled (explicitly).

StevenXL:
> > It is included in 'GHC2021', and thus enabling this extension is consistent with this guide's current recommendation to use 'GHC2021'.
> 
> I don't get this. If it's already included in `GHC2021`, then this guide's recommendation to use `GHC2021` means it should **not** be enabled (explicitly).

A few things. When I read the documentation for GHC2021, I read the "The GHC2021 language edition includes the following extensions:" as a list of extensions it knows about, but not a list of extensions that are enabled by default. So I think you are right, nice catch! 

It looks like the warning I saw which made me try this change is coming from building `shake`, not from our own code. Thank you!

PR #98, opened 2025-03-19
Start haskell persistent guide

chris-martin:
Three topics so far in this guide:

- Use type applications in queries (I think we do this in most places)
- Use OverloadedLabels to refer to entity fields (done in curricula and progress)
- Use unqualified record field names (done only in curricula)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,28 +12,13 @@ Conventions for naming database entities, JSON serializations, etc.
 
 ## Haskell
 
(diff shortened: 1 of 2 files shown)


mjgpy3 (approved):
LGTM

jason-lieb (approved):

OlaoluwaM (approved):
LGTM

